PR #106, opened 2020-09-03
Explain why 2nd test case is valid in GatherSequentialAwaitRule

zsol:
In the generated docs (https://fixit.readthedocs.io/en/latest/rules/GatherSequentialAwaitRule.html), the second valid test case looks the same as the first invalid case, which is confusing. Reported by @abesto

### DIFF
--- a/fixit/rules/gather_sequential_await.py
+++ b/fixit/rules/gather_sequential_await.py
@@ -24,9 +24,10 @@ class GatherSequentialAwaitRule(CstLintRule):
                 return await async_bar()
             """
         ),
-        # await in a loop is fine if it's a test.
         Valid(
             """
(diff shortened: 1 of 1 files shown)


fixit/rules/gather_sequential_await.py
```diff
@@ -24,9 +24,10 @@ async def async_foo():
                 return await async_bar()
             """
         ),
-        # await in a loop is fine if it's a test.
         Valid(
```
jimmylai, line 27: Assuming this is to address the generated documentation issue.
A test filename is provided for this test case.
We can simply also provide the file name in the generated doc to make it clear.

zsol: Yep, that also works!

PR #107, opened 2020-09-03
[docs] add custom path and LintConfig to examples

jimmylai:
## Summary
add custom path and LintConfig to examples
Add path and LintConfig if they are customized in the test cases.

## Test Plan
Before: The examples is confusing since it missed lots of context.
image

After: It's much easy to understand now.
image


josieesh (approved):
Awesome! 😎

PR #108, opened 2020-09-03
Run pyre query initially

josieesh:
## Summary
- So we don't get pyre timeout errors in the tutorials, let's run `pyre query` in the background during start up of the notebook. This way we'll have a cache of types

## Test Plan
Docs before: 
image
Docs after:
image

### DIFF
--- a/docs/source/getting_started.ipynb
+++ b/docs/source/getting_started.ipynb
@@ -79,7 +79,8 @@
     "fp.close()\n",
     "! pyre start --store-type-check-resolution\n",
     "! git init\n",
-    "! git add example.py"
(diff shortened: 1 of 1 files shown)


jimmylai (approved):
Should we add some note to the tutorial to tell developer that they'll need to install `Pyre` and run`Pyre start` if they want to take advantage of Type Inference?
Currently Pyre is listed as a DEV dependency.
So if they run fixit on their repo, they'll see the prompt of "Pyre start" which they'll need to do `pip install pyre-check` if they don't use Pyre before.
https://github.com/Instagram/Fixit/blob/master/requirements-dev.txt

PR #109, opened 2020-09-03
RemoveUnusedSuppressionsRule

josieesh:
## Summary
a special rule that checks for unused lint suppressions
- this rule cannot be included in the `fixit.rules` package as it needs to be run after all other rules. So we call it exclusively from `lint_file` after all other rules have run
- if a lint rule was not run (ie was not included in the set of lint rules passed to `lint_file`, we cannot assume that a comment suppressing this unknown rule is an unused suppression
- for backwards compatibility, and for existing tests to not have to wait on UnusedSuppressionsRule, the default value of `find_unused_suppressions` argument in `lint_file` is `False`

## Test Plan
- unit tests
- Example file `z.py`:
```
l = [1,2,3]
# lint-ignore: NoRedundantFStringRule, NoRedundantListComprehensionRule: test
# lint: test test test test test test
any([i for i in l])
```
run: `python -m fixit.cli.apply_fix z.py` (runs all rules)
see file changed to:
```
l = [1, 2, 3]
# lint-ignore: NoRedundantListComprehensionRule: test test test test test
# lint: test test
any([i for i in l])
```
and reported to user:
image

- [x] used suppression code not removed
- [x] unused suppression code removed
- [x] comment reason left unchanged


fixit/rule_lint_engine.py
```diff
@@ -140,6 +157,7 @@ def lint_file_and_apply_patches(
     rules: LintRuleCollectionT,
     max_iter: int = 100,
     cst_wrapper: Optional[MetadataWrapper] = None,
+    find_unused_suppressions: bool = False,
```
jimmylai, line 160: Why it's False here?

fixit/common/unused_suppressions.py
```diff
@@ -0,0 +1,285 @@
...
+            self._handle_emptyline_sequence_attribute(node, "empty_lines")
+
+    def _handle_suppression_comment(
+        self,
+        parent_node: cst.CSTNode,
+        parent_attribute_name: str,
+        index: int,
+        local_supp_comment: SuppressionComment,
+        comment_physical_line: int,
+    ) -> None:
```
jimmylai, line 217: This looks like something can be refactored to be generic for reuse in LibCST in the future.

jimmylai (approved):

PR #110, opened 2020-09-03
[docs] use yaml format for LintConfig

jimmylai:
## Summary
use yaml format for LintConfig

## Test Plan
image


josieesh (approved):
even more awesome 😎

PR #112, opened 2020-09-03
Implement NoReduntantLambdaFunction

isidentical:
> I am not sure whether this tool itself accepts new rules or not, I just wanted to play with it a little bit so drafted a PR, if the built-in ruleset is intended to kept only for 'selected' rules, it is OK to close this PR.

## Summary
- Implements a lint rule for 'redundant lambda wrapping'.
The pattern mostly happens on function compositions, like `Pool.execute(<func>)` or even `map`. 
```py
def square(x):
    return x ** 2
map(lambda x: square(x), [1,2,3,4])
```
becomes
```py
def square(x):
    return x ** 2
map(square, [1,2,3,4])
```

## Test Plan
I've added both valid and invalid test cases, not sure if more are needed.

### DIFF
--- a/fixit/rules/no_redundant_lambda.py
+++ b/fixit/rules/no_redundant_lambda.py
@@ -1,0 +1,77 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import libcst as cst
+import libcst.matchers as m
+from libcst.helpers import get_full_name_for_node
+
+from fixit import CstLintRule, InvalidTestCase as Invalid, ValidTestCase as Valid
+
+
+UNNECESSARY_LAMBDA: str = (
+    "The lambda that is wrapping {function} is redundant. "
+    + "It can unwrapped safely and used purely."
+)
+
+
+class NoRedundantLambdaFunction(CstLintRule):
+
+    VALID = [
+        Valid("lambda x: foo(y)"),
+        Valid("lambda x: foo(x, y)"),
+        Valid("lambda x, y: foo(x)"),
+        Valid("lambda x, /: foo(x)"),
+        Valid("lambda *, x: foo(x)"),
+        Valid("lambda x = y: foo(x)"),
+        Valid("lambda x, y: foo(y, x)"),
+        Valid("lambda x, y: foo(y=x, x=y)"),
+        Valid("lambda x, y, *z: foo(x, y, z)"),
+        Valid("lambda x, y, **z: foo(x, y, z)"),
+    ]
+    INVALID = [
+        Invalid("lambda x: foo(x)", expected_replacement="foo"),
+        Invalid(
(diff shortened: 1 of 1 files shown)


jimmylai:
@isidentical new rules are definite welcome, especially those generic rules that can make code cleaner, safer or more efficient.

fixit/rules/no_redundant_lambda.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+UNNECESSARY_LAMBDA: str = (
+    "The lambda that is wrapping {function} is redundant. "
+    + "It can unwrapped safely and used purely."
+)
+
+
+class NoRedundantLambdaFunction(CstLintRule):
+
```
jimmylai, line 20: We use Rule as the naming convention.
https://fixit.readthedocs.io/en/latest/build_a_lint_rule.html#Pick-a-Good-Name
```suggestion
class NoRedundantLambdaRule(CstLintRule):

```

fixit/rules/no_redundant_lambda.py
```diff
@@ -0,0 +1,76 @@
...
+from fixit import CstLintRule, InvalidTestCase as Invalid, ValidTestCase as Valid
+
+
+UNNECESSARY_LAMBDA: str = (
+    "The lambda that is wrapping {function} is redundant. "
+    + "It can unwrapped safely and used purely."
+)
+
+
+class NoRedundantLambdaFunction(CstLintRule):
```
jimmylai, line 19: Can you add some more context as docsstring of this class?

The docstring will be included in the generated doc. We require some description to explain each rule.

https://fixit.readthedocs.io/en/latest/rules/NoInheritFromObjectRule.html

fixit/rules/no_redundant_lambda.py
```diff
@@ -0,0 +1,76 @@
...
+        Valid("lambda x, y: foo(y=x, x=y)"),
+        Valid("lambda x, y, *z: foo(x, y, z)"),
+        Valid("lambda x, y, **z: foo(x, y, z)"),
+    ]
+    INVALID = [
+        Invalid("lambda x: foo(x)", expected_replacement="foo"),
+        Invalid(
+            "lambda x, y, z: (t + u).math_call(x, y, z)",
+            expected_replacement="(t + u).math_call",
+        ),
```
jimmylai, line 37: Let's add some edge case example.

E.g. lambda has no arguments like `lambda: self.func()`.

jimmylai (changes requested):
Thanks for contributing! This is a very useful autofixer which can make the code simpler and more efficient!
I was able to easily find hundreds of violations in our codebase.
Just some minor change requests.

fixit/rules/no_redundant_lambda.py
isidentical: Oh completely missed that tutorial! Thanks

jimmylai:
Looks awesome!

isidentical:
> Looks awesome!
Wow, Thanks!

PR #116, opened 2020-09-04
Implement ChainedInstanceRule

isidentical:
## Summary
```
if isinstance(x, int) or isinstance(x, str): ...
elif isinstance(x, bytes) or isinstance(x, bytearray): ...
```
becomes
```
if isinstance(x, (int, str)): ...
elif isinstance(x, (bytes, bytearray)): ...
```

## Test Plan
- Added unittests

### DIFF
--- a/fixit/rules/chained_instance_check.py
+++ b/fixit/rules/chained_instance_check.py
@@ -16,7 +16,7 @@ from fixit import (
 )
 
 
-class ChainedInstanceCheckRule(CstLintRule):
+class CollapseIsinstanceChecksRule(CstLintRule):
     """
     The built-in ``isinstance`` function, instead of a single type,
     can take a tuple of types and check whether given target suits
--- a/fixit/rules/chained_instance_check.py
+++ b/fixit/rules/chained_instance_check.py
@@ -32,8 +32,10 @@ class ChainedInstanceCheckRule(CstLintRule):
     )
 
     VALID = [
+        Valid("foo() or foo()"),
         Valid("foo(x, y) or foo(x, z)"),
         Valid("foo(x, y) or foo(x, z) or foo(x, q)"),
+        Valid("isinstance() or isinstance()"),
         Valid("isinstance(x) or isinstance(x)"),
         Valid("isinstance(x, y) or isinstance(x)"),
         Valid("isinstance(x) or isinstance(x, y)"),
--- a/fixit/rules/chained_instance_check.py
+++ b/fixit/rules/chained_instance_check.py
@@ -3,15 +3,20 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import List, Set
+from typing import List, Optional, Set
 
 import libcst as cst
 import libcst.matchers as m
 
-from fixit import CstLintRule, InvalidTestCase as Invalid, ValidTestCase as Valid
+from fixit import (
+    CstContext,
+    CstLintRule,
+    InvalidTestCase as Invalid,
+    ValidTestCase as Valid,
+)
 
 
-class ChainedInstanceRule(CstLintRule):
+class ChainedInstanceCheckRule(CstLintRule):
     """
     The built-in ``isinstance`` function instead of a single type,
     can take a tuple of types and check whether given target suits
@@ -52,8 +57,8 @@ class ChainedInstanceRule(CstLintRule):
         ),
     ]
 
-    def __init__(self, *args, **kwargs) -> None:
-        super().__init__(*args, **kwargs)
+    def __init__(self, context: CstContext) -> None:
+        super().__init__(context)
 
         # Since we already unwrap a boolean op's
         # children
@@ -69,18 +74,23 @@ class ChainedInstanceRule(CstLintRule):
         ):
             args = self._collect_isinstance_args(node)
             if args is not None:
+                right = cst.ensure_type(node.right, cst.Call)
                 elements = [cst.Element(arg.value) for arg in args]
-                new_isinstance_call = node.right.with_deep_changes(
-                    old_node=node.right.args[1], value=cst.Tuple(elements)
+                new_isinstance_call = right.with_deep_changes(
+                    old_node=right.args[1], value=cst.Tuple(elements)
                 )
                 self.report(node, replacement=new_isinstance_call)
-                return False
 
-    def _collect_isinstance_args(self, node: cst.BooleanOperation) -> List[cst.Arg]:
+    def _collect_isinstance_args(
+        self, node: cst.BooleanOperation
+    ) -> Optional[List[cst.Arg]]:
         target = cst.ensure_type(node.right, cst.Call).args[0].value
         expected_call = m.Call(
             func=m.Name(value="isinstance"),
-            args=[m.Arg(value=target), m.Arg(value=~m.Tuple())],
+            args=[
+                m.Arg(value=m.MatchIfTrue(target.deep_equals)),
+                m.Arg(value=~m.Tuple()),
+            ],
         )
         expected_boolop = m.BooleanOperation(operator=m.Or(), right=expected_call)
 
@@ -88,6 +98,7 @@ class ChainedInstanceRule(CstLintRule):
         stack = []
         current = node
         while m.matches(current, expected_boolop):
+            current = cst.ensure_type(current, cst.BooleanOperation)
             seen.append(current)
             stack.insert(0, current.right)
             current = current.left
--- a/fixit/rules/chained_instance_check.py
+++ b/fixit/rules/chained_instance_check.py
@@ -18,17 +18,17 @@ from fixit import (
 
 class ChainedInstanceCheckRule(CstLintRule):
     """
-    The built-in ``isinstance`` function instead of a single type,
+    The built-in ``isinstance`` function, instead of a single type,
     can take a tuple of types and check whether given target suits
-    any of them. Instead of chaining multiple ``isinstance`` calls
-    with a boolean-or operation, they can be simplified into a single
-    ``isinstance`` call.
+    any of them. Rather than chaining multiple ``isinstance`` calls
+    with a boolean-or operation, a single ``isinstance`` call where
+    the second argument is a tuple of all types can be used.
     """
 
     MESSAGE: str = (
         "Multiple isinstance calls with the same target but "
         + "different types can be collapsed into a single call "
-        + " with a tuple of types"
+        + "with a tuple of types."
     )
 
     VALID = [
--- a/fixit/rules/chained_instance_check.py
+++ b/fixit/rules/chained_instance_check.py
@@ -73,7 +73,7 @@ class CollapseIsinstanceChecksRule(CstLintRule):
         ),
         Invalid(
             "something or isinstance(x, y) or isinstance(x, z) or another",
-            expected_replacement="something or isinstance(x, (y, z)) or another"
+            expected_replacement="something or isinstance(x, (y, z)) or another",
         ),
         Invalid(
             "isinstance(x, y) or isinstance(x, z) or isinstance(x, q) or isinstance(x, w)",
--- a/fixit/rules/chained_instance_check.py
+++ b/fixit/rules/chained_instance_check.py
@@ -3,16 +3,22 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import List, Optional, Set
+from typing import List, Optional, Set, Union
 
 import libcst as cst
 import libcst.matchers as m
+from libcst.metadata import QualifiedName, QualifiedNameProvider, QualifiedNameSource
 
 from fixit import (
     CstContext,
     CstLintRule,
     InvalidTestCase as Invalid,
     ValidTestCase as Valid,
+)
+
+
+_ISINSTANCE = QualifiedName(
+    name="builtins.isinstance", source=QualifiedNameSource.BUILTIN
 )
 
 
@@ -31,6 +37,8 @@ class CollapseIsinstanceChecksRule(CstLintRule):
         + "with a tuple of types."
     )
 
+    METADATA_DEPENDENCIES = (QualifiedNameProvider,)
+
     VALID = [
         Valid("foo() or foo()"),
         Valid("foo(x, y) or foo(x, z)"),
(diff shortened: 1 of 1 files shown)


zsol (commented):
Looks like a great lint rule, thanks for working on it. I only have one comment.

fixit/rules/chained_instance_check.py
```diff
@@ -0,0 +1,112 @@
...
+    VALID = [
+        Valid("foo(x, y) or foo(x, z)"),
+        Valid("foo(x, y) or foo(x, z) or foo(x, q)"),
+        Valid("isinstance(x) or isinstance(x)"),
+        Valid("isinstance(x, y) or isinstance(x)"),
+        Valid("isinstance(x) or isinstance(x, y)"),
+        Valid("isinstance(x, y) or isinstance(t, y)"),
+        Valid("isinstance(x, y) and isinstance(x, z)"),
+        Valid("isinstance(x, y) or isinstance(x, (z, q))"),
+        Valid("isinstance(x, (y, z)) or isinstance(x, q)"),
```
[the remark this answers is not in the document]
isidentical: In my opinion, no. Since the developer is aware that they can use a tuple of types in `isinstance` calls, the best estimation would be that they wanted to separate a check into two different groups for readability.
```py
if isinstance(x, (int, float, complex)) or isinstance(x, DateT): # either numeric or date
     ...
```

fixit/rules/chained_instance_check.py
```diff
@@ -0,0 +1,112 @@
...
+                elements = [cst.Element(arg.value) for arg in args]
+                new_isinstance_call = right.with_deep_changes(
+                    old_node=right.args[1], value=cst.Tuple(elements)
+                )
+                self.report(node, replacement=new_isinstance_call)
+
+    def _collect_isinstance_args(
+        self, node: cst.BooleanOperation
+    ) -> Optional[List[cst.Arg]]:
+        target = cst.ensure_type(node.right, cst.Call).args[0].value
```
zsol, line 87: This relies on `args` being nonempty, and while that's guaranteed by the matcher on line 73, best to add a test case for it. Something like
`Valid(foo() or foo())` or even `Valid(isinstance() or isinstance())`

fixit/rules/chained_instance_check.py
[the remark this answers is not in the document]
zsol: Fair enough.

zsol (approved):
LGTM, awesome work :)

isidentical:
> LGTM, awesome work :)

Glad to hear that, thanks!

fixit/rules/chained_instance_check.py
```diff
@@ -0,0 +1,114 @@
...
+
+from fixit import (
+    CstContext,
+    CstLintRule,
+    InvalidTestCase as Invalid,
+    ValidTestCase as Valid,
+)
+
+
+class ChainedInstanceCheckRule(CstLintRule):
```
jimmylai, line 19: We hope the lint rule name is short and actionable. So developer can learn the action to take easily.
https://fixit.readthedocs.io/en/latest/build_a_lint_rule.html#Pick-a-Good-Name

What do you think about rename as`CollapseIsinstanceChecksRule`?
```suggestion
class CollapseIsinstanceChecksRule(CstLintRule):
```

fixit/rules/chained_instance_check.py
```diff
@@ -0,0 +1,114 @@
...
+                    old_node=right.args[1], value=cst.Tuple(elements)
+                )
+                self.report(node, replacement=new_isinstance_call)
+
+    def _collect_isinstance_args(
+        self, node: cst.BooleanOperation
+    ) -> Optional[List[cst.Arg]]:
+        target = cst.ensure_type(node.right, cst.Call).args[0].value
+        expected_call = m.Call(
+            func=m.Name(value="isinstance"),
```
jimmylai, line 91: To make this rule more robust, we'll want to make sure the `isinstance` is `builtins.isinstance` rather than a local defined function or an import alias.
We can use `QualifiedNameProvider.has_name` to check this.
Add a metadata dependency `METADATA_DEPENDENCIES = (QualifiedNameProvider,)` to the lint rule and then 
use the has_name helper to check it, e.g.
https://github.com/Instagram/Fixit/blob/18365db024258817f7f30006886c14ce8f07bbd5/fixit/rules/cls_in_classmethod.py#L225-L230

Reference for metadata API.
https://libcst.readthedocs.io/en/latest/metadata_tutorial.html
https://libcst.readthedocs.io/en/latest/metadata.html

fixit/rules/chained_instance_check.py
```diff
@@ -0,0 +1,114 @@
...
+            expected_replacement="isinstance(x, (y, z))",
+        ),
+        Invalid(
+            "isinstance(x, y) or isinstance(x, z) or isinstance(x, q)",
+            expected_replacement="isinstance(x, (y, z, q))",
+        ),
+        Invalid(
+            "isinstance(x, y) or isinstance(x, z) or isinstance(x, q) or isinstance(x, w)",
+            expected_replacement="isinstance(x, (y, z, q, w))",
+        ),
```
jimmylai, line 59: Can you add an edge case that multiple `isinstance` were used and they belong to different groups, each group is collapsed correctly?

E.g.
`isinstance(x, a) or isinstance(x, b) or isinstance(y, c) or isinstance(y, d)` is fixed as
`isinstance(x, (a, b)) or isinstance(y, (c, d))`

jimmylai (commented):
@isidentical , this is a great rule that makes code more efficient and more simpler.
I've used this to optimize some hot path at work.

The code is 100% tested. Awesome job!
https://codecov.io/gh/Instagram/Fixit/pull/116/diff?src=pr&el=tree#diff-Zml4aXQvcnVsZXMvY2hhaW5lZF9pbnN0YW5jZV9jaGVjay5weQ==

Other than `isinstance`, some other Python builtin methods can also take a Tuple to combine multiple checks e.g. `startswith` and `endswith` of str. The pattern is slightly different (it needs to be a `str.startswith`, so probably better to build as another lint rule).

isidentical:
> Other than isinstance, some other Python builtin methods can also take a Tuple to combine multiple checks e.g. startswith and endswith of str. The pattern is slightly different (it needs to be a str.startswith, so probably better to build as another lint rule).

Exactly, that is what I was thinking, you basically read my mind!

fixit/rules/chained_instance_check.py
isidentical: `Isinstance` doesn't look so nice (personally would prefer `IsInstance`, but not a big deal), but OK, sure!

fixit/rules/chained_instance_check.py
isidentical: Yeah, makes sense!

fixit/rules/chained_instance_check.py
isidentical: I had to alter the implementation for supporting multiple-target changes, here it comes.

fixit/rules/chained_instance_check.py
```diff
@@ -0,0 +1,188 @@
...
+
+        targets = {}
+        for call in stack:
+            func = cst.ensure_type(call, cst.Call).func
+            for context in self.get_metadata(QualifiedNameProvider, func):
+                if (
+                    context.name != "builtins.isinstance"
+                    or context.source is not QualifiedNameSource.BUILTIN
+                ):
+                    return None
```
jimmylai, line 161: Why not simply use `self.QualifiedNameProvider.has_name(name="builtins.isinstance", QualifiedNameSource.BUILTIN)` to be simpler?

jimmylai (commented):
CollapseIsInstanceChecksRule (with capital `I`) also looks good to me if you like it more.

fixit/rules/chained_instance_check.py
jimmylai: Also, this checks looks redundant since we already check `isinstance` in `_collect_isinstance_calls` and we can simply use `QualifiedNameProvider.has_name(self, name="builtins.isinstance", QualifiedNameSource.BUILTIN)` in `_collect_isinstance_calls`.

fixit/rules/chained_instance_check.py
```diff
@@ -0,0 +1,188 @@
...
+            if isinstance(x, y) or isinstance(x, z):
+                pass
+            """
+        ),
+    ]
+    INVALID = [
+        Invalid(
+            "isinstance(x, y) or isinstance(x, z)",
+            expected_replacement="isinstance(x, (y, z))",
+        ),
```
jimmylai, line 68: The current implementation doesn't work for cases combined with other operator, e.g.
"something and isinstance(x, a) or isinstance(x, b)"
"something or isinstance(x, a) or isinstance(x, b)".


To make the implementation right, we can probably keep track of valid isinstance checks in `visit_BooleanOperation` and then try to generate the collapsed from bottom up recursion in `visit_BooleanOperation`. That way, each `visit_BooleanOperation` step tries to collapse two isintance checks and will only report the the replacement when no more collapse-able operator can be found.

jimmylai (changes requested):

fixit/rules/chained_instance_check.py
isidentical: AFAIK `has_name` check for any appearance, but we want to ensure that it is always the `isinstance` we found (just like `all` instead of `any`). I don't know though if it is the right way or not, can change?

fixit/rules/chained_instance_check.py
isidentical: I am a bit confused on this example `something and isinstance(x, a) or isinstance(x, b)`, since it shouldn't be collapsed (precedence of AND > OR) . For this one, will try to find a solution `something or isinstance(x, a) or isinstance(x, b).`

jimmylai: You're right, considering the precedence, it cannot be collapsed.

isidentical: I think I found a very simple and straight forward way of building multi-item or-expressions, but still needs much tests. Short preview;
```
        def unwrap(node: cst.BaseExpression):
            if m.matches(node, m.BooleanOperation(operator=cst.Or())):
                yield from unwrap(node.left)
                yield node.right
            else:
                yield node
        
        names = {}
        stack = tuple(unwrap(node))
...
```

fixit/rules/chained_instance_check.py
jimmylai: Yes, `has_name` checks if any of the qualified name matches and that's how we usually checks the qualified names. https://libcst.readthedocs.io/en/latest/metadata.html#libcst.metadata.QualifiedNameProvider

The `QualifiedName` analysis is based on scope analysis to find assignments from parent scopes and it returns all possible assignments. Sometimes when the code has branches with redefinition of the same name, it could return multiple qualified names. Use `all` can miss lint suggestion on those cases while use `any` can provide the correct suggestion based on our experience.
E.g.
https://github.com/Instagram/LibCST/blob/7ca738bf3964ac48ebb36b1b362153c87c29a706/libcst/codemod/commands/rename.py#L262-L267
https://github.com/Instagram/LibCST/blob/c023fa7c4caff3fd2b3946080f9a58b539b10363/libcst/codemod/commands/convert_namedtuple_to_dataclass.py#L31-L47

fixit/rules/chained_instance_check.py
```diff
@@ -0,0 +1,173 @@
...
+        Valid("isinstance(x, y) and isinstance(x, z)"),
+        Valid("isinstance(x, y) or isinstance(x, (z, q))"),
+        Valid("isinstance(x, (y, z)) or isinstance(x, q)"),
+        Valid("isinstance(x, a) or isinstance(y, b) or isinstance(z, c)"),
+        Valid(
+            """
+            def foo():
+                def isinstance(x, y):
+                    return _foo_bar(x, y)
+                if isinstance(x, y) or isinstance(x, z):
```
jimmylai, line 60: Why this is not collapsed? This is basically the same as the first Invalid test case and should be collapsed.

isidentical: Did you see the function `def isinstance(x, y):` that redefines `isinstance` name in the current scope?

jimmylai: I missed that. It makes sense now!

jimmylai (approved):
Looks great!
Currently the rule only consider the bare isinstance and or operator. In some other collapse-able case like `not isinstance(...) and not isinstance(...)` which can be collapsed as `not isinstance(..., [...])`. I think that's a less common use case and can leave as a future enhancement if you like to improve it.

Found another builtin function `issubclass` can also takes a Tuple but it's an `all` check (unlike the `any` check in isinstance). So `issubclass(x, y) and issubclass(x, z)` can be collapsed. That can be another lint rule by reusing the unwrap and collect_targets helpers provided in this lint rule.

isidentical:
> Found another builtin function issubclass can also takes a Tuple but it's an all check (unlike the any check in isinstance). So issubclass(x, y) and issubclass(x, z) can be collapsed. That can be another lint rule by reusing the unwrap and collect_targets helpers provided in this lint rule.

I am aware of `issubclass` taking a tuple of classes, but AFAIK it is also an or gate. `issubclass(x, (y, z)) == issubclass(x, y) or issubclass(x, z)`

jimmylai:
> > Found another builtin function issubclass can also takes a Tuple but it's an all check (unlike the any check in isinstance). So issubclass(x, y) and issubclass(x, z) can be collapsed. That can be another lint rule by reusing the unwrap and collect_targets helpers provided in this lint rule.
> 
> I am aware of `issubclass` taking a tuple of classes, but AFAIK it is also an or gate. `issubclass(x, (y, z)) == issubclass(x, y) or issubclass(x, z)`
Just confirmed that it's an `OR`. I was confused by the documentation in 
https://docs.python.org/3/library/functions.html#issubclass
It says "in which case every entry in classinfo will be checked". If it'll return early when any subclass matches (which it should do to be efficient), then the description isn't fully accurate.

@isidentical would you like to refactor `CollapseIsinstanceChecksRule` to provide collapsing issubclass checks in the same rule or as another rule?

isidentical:
> @isidentical would you like to refactor CollapseIsinstanceChecksRule to provide collapsing issubclass checks in the same rule or as another rule?

I think it suits to the purpose, so I would be OK with it. No need to mirroring the whole logic. If it sounds nice to you, I can just send a very small / simple PR for that.

jimmylai:
> > @isidentical would you like to refactor CollapseIsinstanceChecksRule to provide collapsing issubclass checks in the same rule or as another rule?
> 
> I think it suits to the purpose, so I would be OK with it. No need to mirroring the whole logic. If it sounds nice to you, I can just send a very small / simple PR for that.

Yes, please go head, I'm happy to review and merge it.
Duplicate the code is not ideal.

PR #120, opened 2020-09-06
Replace fstring url

### LINKED ISSUE #115: Use of old-style format leads to advice with broken link
Hey,

Thanks for the tool. 

There appears to be an invalid link in (at least) one of the advices.

Note the attached file. 

When running `python -m fixit.cli.run_rules ` with a file with the following contents:
```
myname = "richie"
string_old = "This is {name}'s string".format(name=myname)

string_new = f"This is {myname}'s string"
```

I get the following advice:

> invalid_link_suggestion.py:2:14
>    UseFstringRule: Do not use printf style formatting or .format(). Use f-string instead to be more readable and
>   efficient. See https://fburl.com/usefstring.```

I mean, I like fburls, I remember them well. However, I (and pretty much every non-FB employee) have no access to them. 

It would probably be better if this pointed to a link accessible from the public web.

acharles7:
## Summary

Fixes #115
* Replaced  https://fburl.com/usefstring -> https://www.python.org/dev/peps/pep-0498/ 
* Added `.idea/` to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ build/
 docs/source/rules/
 coverage.xml
 coverage_report.html
+.idea/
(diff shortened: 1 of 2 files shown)


jimmylai (approved):
LGTM

Thanks for contributing! @acharles7

PR #123, opened 2020-09-08
Script to add new rule file

### LINKED ISSUE #119: Script to add new rules
I think it would be a great idea to add a script to generate a higher-level skeleton of adding a new rule in [fixit/rules](https://github.com/Instagram/Fixit/tree/master/fixit/rules)
The possible skeleton would be
```python 

# Copyright (c) Facebook, Inc. and its affiliates.
#
# This source code is licensed under the MIT license found in the
# LICENSE file in the root directory of this source tree.

import libcst as cst

from fixit import CstLintRule, InvalidTestCase as Invalid, ValidTestCase as Valid

class NewRule(CstLintRule):
    '''
    docstring or new_rule information
    '''
    MESSAGE = 'message'
    VALID = [Valid()]
    INVALID = [Invalid()]
```

Right now, I have only added license, imports, and class. But, we can add other things too if needed in future.
if we generate this skeleton using python scripts then It will be very easy for other contributors to add new rules.

So, to add new rule, run `add_new_rule.py` file (will be located in fixit/scripts) using
`python -m fixit.scripts.add_new_rule --name rule_name`
This will add new_rule.py in [fixit/rules](https://github.com/Instagram/Fixit/tree/master/fixit/rules)

Please let me know what you think.
I am happy to make PR for this script in fixit package.

acharles7:
## Summary

Fixes #119 
* Added script to create `new_rule.py`
```bash
$ python -m fixit.cli.add_new_rule --path fixit/rules/new_rule.py
```

This will create a skeleton of the rule file at the path given in the `--path` argument

`new_rule.py`
```python
# Copyright (c) Facebook, Inc. and its affiliates.
#
# This source code is licensed under the MIT license found in the
# LICENSE file in the root directory of this source tree.


from fixit import CstLintRule, InvalidTestCase as Invalid, ValidTestCase as Valid


"""
This is a model rule file for adding a new rule to fixit module
"""


class Rule(CstLintRule):
    """
    docstring or new_rule description
    """

    MESSAGE = "Enter rule description message"

    VALID = [Valid("'example'")]

    INVALID = [Invalid("'example'")]
```


fixit/cli/add_new_rule.py
```diff
@@ -0,0 +1,88 @@
...
+        raise FileNotFoundError(f"{filepath} already exists")
+    elif not filepath.parent.exists():
+        raise TypeError(f"{filepath} is not a valid path, provide path with file name")
+    else:
+        return filepath
+
+
+def create_rule_file(file_path: Path) -> None:
+    """Create a new rule file."""
+    context = _LICENCE + _IMPORTS + _TO_DOS + _RULE_CLASS
```
jimmylai, line 64: Let's use the formatter to reformat the code based on user's config.
https://github.com/Instagram/Fixit/blob/ef5edca7c6d3cc5f2f1cb5ff54b769ac8bb0d923/fixit/cli/apply_fix.py#L126-L128

jimmylai (approved):
Let's also include this tool in the documentation.

https://github.com/Instagram/Fixit/blob/46a0d0b66e494c7dad4fe5d57dea9981a96df322/docs/source/build_a_lint_rule.ipynb

You can run `jupyter notebook` in your local venv to start Jupyter notebook and edit it in browser.
It uses nbsphinx to generate doc and text cell needs to use `rst` format.
https://nbsphinx.readthedocs.io/en/0.7.1/raw-cells.html
The best place to mention this command is in the section of "Lint Rule Scaffolding".

fixit/cli/add_new_rule.py
```diff
@@ -0,0 +1,88 @@
...
+
+"""
+Usage:
+   $ python -m fixit.cli.add_new_rule --help
+   $ python -m fixit.cli.add_new_rule
+   $ python -m fixit.cli.add_new_rule --path fixit/rules/new_rule.py
+
+"""
+
+_LICENCE = """# Copyright (c) Facebook, Inc. and its affiliates.
```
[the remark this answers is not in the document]
acharles7:
```
_LICENCE = """\
# Copyright (c) Facebook, Inc. and its affiliates.
```
this gives me lint error in windows python that's why I removed it.
Please suggest me other way to do this.

fixit/cli/add_new_rule.py
```diff
@@ -0,0 +1,93 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in thefla
```
isidentical, line 3:
```suggestion
# This source code is licensed under the MIT license found in the
```

fixit/cli/add_new_rule.py
```diff
@@ -0,0 +1,93 @@
...
+
+_LICENCE = """\
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+
+_IMPORTS = """
```
isidentical, line 28: Just my opinion, but wouldn't it be cool to have `libcst as cst` and `libcst.matchers as m` imports here as well?

isidentical:
As far as I could estimate from writing 2 rules is that, saving the rule to `new_file.py` by default isn't that useful. Maybe it can take a positional argument named rule name and use it on both places (the filename and the class name, so that people wouldn't need to replace the class name, because they might just remove the `Rule` at all which is non-compliant with the style standards of this repo) (just my opinion)

acharles7:
> As far as I could estimate from writing 2 rules is that, saving the rule to `new_file.py` by default isn't that useful. Maybe it can take a positional argument named rule name and use it on both places (the filename and the class name, so that people wouldn't need to replace the class name, because they might just remove the `Rule` at all which is non-compliant with the style standards of this repo) (just my opinion)

I already suggest this, but As per @jimmylai suggestion, I have not implemented this functionality.

fixit/cli/add_new_rule.py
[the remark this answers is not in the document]
isidentical: It is weird that only windows fail, and I have no idea. Guess you have to revert this.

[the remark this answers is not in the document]
acharles7: I'll revert it.

fixit/cli/add_new_rule.py
acharles7: Will add those imports.

fixit/cli/add_new_rule.py
```diff
@@ -0,0 +1,88 @@
...
+from pathlib import Path
+
+
+"""
+Usage:
+   $ python -m fixit.cli.add_new_rule --help
+   $ python -m fixit.cli.add_new_rule
+   $ python -m fixit.cli.add_new_rule --path fixit/rules/new_rule.py
+
+"""
```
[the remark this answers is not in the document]
acharles7: Done

fixit/cli/add_new_rule.py
acharles7: Done

fixit/cli/add_new_rule.py
acharles7: Done

fixit/cli/add_new_rule.py
acharles7: Done

acharles7:
> > As far as I could estimate from writing 2 rules is that, saving the rule to `new_file.py` by default isn't that useful. Maybe it can take a positional argument named rule name and use it on both places (the filename and the class name, so that people wouldn't need to replace the class name, because they might just remove the `Rule` at all which is non-compliant with the style standards of this repo) (just my opinion)
> 
> I already suggest this, but As per @jimmylai suggestion, I have not implemented this functionality.

Will add `--name abcd` later on for
* create filename, `abcd.py`
* class name, `class AbcdRule`

fixit/cli/add_new_rule.py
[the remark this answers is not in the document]
jimmylai: Yeah, I saw the error on Windows before and didn't know how to fix it.

jimmylai:
> > As far as I could estimate from writing 2 rules is that, saving the rule to `new_file.py` by default isn't that useful. Maybe it can take a positional argument named rule name and use it on both places (the filename and the class name, so that people wouldn't need to replace the class name, because they might just remove the `Rule` at all which is non-compliant with the style standards of this repo) (just my opinion)
> 
> I already suggest this, but As per @jimmylai suggestion, I have not implemented this functionality.

The class name is in camel case and the file name is in underscore-separated-case. They doesn't necessary to be the same name. So provide another parameter like `--rule` or `--name` for class name makes more sense to me.

isidentical:
> The class name is in camel case and the file name is in underscore-separated-case. They doesn't necessary to be the same name. So provide another parameter like --rule or --name for class name makes more sense to me.

Yeah, makes sense (the thing that was in my mind was .split('_') + .title() + .join() + 'Rule')

docs/source/build_a_lint_rule.ipynb
```diff
@@ -121,14 +121,86 @@
     "        ..."
    ]
   },
+  {
+   "cell_type": "raw",
+   "metadata": {},
```
jimmylai, line 126: You need to follow the steps in https://nbsphinx.readthedocs.io/en/0.7.1/raw-cells.html to specify the format as `rst`.
(which generates `"raw_mimetype": "text/restructuredtext"` metadata)
Otherwise the output is in wrong format.
image
 
You can run `tox -e docs` locally to generate html files and verify the result in a browser.

jimmylai (changes requested):

docs/source/build_a_lint_rule.ipynb
acharles7: thank you, looking into it

docs/source/build_a_lint_rule.ipynb
```diff
@@ -121,14 +121,86 @@
...
+   "outputs": [],
+   "source": [
+    "# check if any of the base classes of this class def is \"object\"\n",
+    "def visit_ClassDef(self, node: cst.ClassDef):\n",
+    "    has_object_base = any(\n",
+    "        isinstance(arg.value, cst.Name) and arg.value.value == \"object\"\n",
+    "        for arg in node.bases\n",
+    "    )"
+   ]
+  },
```
jimmylai, line 196: Those two cells are repeated.

acharles7:
> > The class name is in camel case and the file name is in underscore-separated-case. They doesn't necessary to be the same name. So provide another parameter like --rule or --name for class name makes more sense to me.
> 
> Yeah, makes sense (the thing that was in my mind was .split('_') + .title() + .join() + 'Rule')

How about this?
```python
def snake_to_camelcase(name):
  """Convert snake-case string to camel-case string."""
  return "".join(n.capitalize() for n in name.split("_"))
```

jimmylai:
> > > The class name is in camel case and the file name is in underscore-separated-case. They doesn't necessary to be the same name. So provide another parameter like --rule or --name for class name makes more sense to me.
> > 
> > 
> > Yeah, makes sense (the thing that was in my mind was .split('_') + .title() + .join() + 'Rule')
> 
> How about this?
> 
> ```python
> def snake_to_camelcase(name):
>   """Convert snake-case string to camel-case string."""
>   return "".join(n.capitalize() for n in name.split("_"))
> ```

Yeah, that works. Then we should explain it clearly in the documentation.

docs/source/build_a_lint_rule.ipynb
acharles7: Removed.

docs/source/build_a_lint_rule.ipynb
acharles7: Format specified

acharles7:
@jimmylai Should I start working on `--name` or Is it for the future?

docs/source/build_a_lint_rule.ipynb
```diff
@@ -211,7 +183,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 5,
```
jimmylai, line 186: `execution_count` is assigned at run time, the value should be `null`.

jimmylai: When any of the cell is already executed, NBSphinx won't run the notebook in CI.
image

acharles7: Okay, will make sure, NBSphinx new to me, that's why. Thank you for helping me with this.

acharles7: done

docs/source/build_a_lint_rule.ipynb
```diff
@@ -130,6 +130,50 @@
...
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
+   },
+   "source": [
+    "Use above structure to create a new rule. \n",
+    "\n",
```
jimmylai, line 176: This looks good now.
Sorry for not thinking of it earlier. Do you mind make one more change?
The best way to demonstrate the command is to actually run it in the tutorial instead of hardcoded the result. So the output will be automatically generated and updated whenever the source code changes.

Can you update this doc to actually run the command to add a new rule?
You can run the command and just cat its content by.
```
! python -m fixit.cli.add_new_rule --path fixit/rules/my_rule.py
! cat fixit/rules/my_rule.py
```
https://github.com/Instagram/Fixit/blob/master/docs/source/getting_started.ipynb
https://fixit.readthedocs.io/en/latest/getting_started.html

acharles7: No worries, I'll add it.

jimmylai:
> @jimmylai Should I start working on `--name` or Is it for the future?

We can leave the `name` parameter change as the next PR and keep this PR simpler.

docs/source/build_a_lint_rule.ipynb
```diff
@@ -130,6 +130,50 @@
...
+    "    \"\"\"\n",
+    "\n",
+    "    MESSAGE = \"Enter rule description message\"\n",
+    "\n",
+    "    VALID = [Valid(\"'example'\")]     # Valid examples\n",
+    "\n",
+    "    INVALID = [Invalid(\"'example'\")] # Invalid examples\n",
+    "    ..."
+   ]
+  },
```
jimmylai, line 168: The changes you added to Getting Started should just replace this part. So whenever lint rule template changes, we don't need to manually update this example.
Getting Started is not need to be updated.

Before commit change, please click on "Restart & Clear Output". So all output are not stored.
The CI job will run through each cell and generate the result on the fly.
image

acharles7: Got it. I thought, we have to run those cells in order to show output.

fixit/cli/add_new_rule.py
```diff
@@ -0,0 +1,95 @@
...
+
+    INVALID = [Invalid("'example'")]
+"""
+
+
+def is_path_exists(path: str) -> Path:
+    """Check for valid path, if yes, return `Path` else raise `Error` """
+    filepath = Path(path)
+    if filepath.exists():
+        raise FileNotFoundError(f"{filepath} already exists")
```
jimmylai, line 60: FileExistsError is better aligned with the error message.
```suggestion
        raise FileExistsError(f"{filepath} already exists")
```

jimmylai:
It's getting closer.
The added commands doesn't work and show this error.
image

It's because the working directory is `docs/source`, rather than the root of project, so it cannot find `fixit/rules`.

Let's change the working directory with a hidden cell.

image

To hide the `os.chdir` cell, you can click View to enable the Edit Metadata button and then add "nbsphinx": "hidden" to the metadata to hide it. So it won't be included in the generated html.
image

Previously, the `add_rule` command is in the same cell as the `cat` which their output will print together.
Let's also separate them as two different cell to be clear.
Lastly, the documentation you added is better for "build_a_lint_rule". Add it to "getting_started" looks redundant. So let's not to add it to "getting_started".

jimmylai:
@acharles7 it looks really awesome now!
Thanks for your contribution and sorry for so many back and forth.

image

acharles7:
@jimmylai Thank you for your quick reviews. 
Should I start working on `--name`?

jimmylai:
> @jimmylai Thank you for your quick reviews.
> Should I start working on `--name`?

Sure, please go ahead.

Issue #124, opened 2020-09-09
Expose fixit as console script


gaborbernat:
Otherwise you can't install it via pipx.

```
pipx install fixit                                                                                                                                                                                                                           Note: Dependent package 'flake8>=3.8.1' contains 1 apps
  - flake8
Note: Dependent package 'pyflakes<2.3.0,>=2.2.0' contains 1 apps
  - pyflakes
Note: Dependent package 'pycodestyle<2.7.0,>=2.6.0a1' contains 1 apps
  - pycodestyle

No apps associated with package fixit. Try again with '--include-deps' to include apps of dependent packages, which are listed above. If you are attempting to install a library, pipx should not be used. Consider using pip or a similar tool instead.
```

jimmylai:
@gaborbernat Nice proposal.
Feel free to add the support for pipx.
It can be done by adding `scripts` or `console_scripts` to `setup.py`.
https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#command-line-scripts

All Fixit scripts are in `fixit/cli/` dir. See https://fixit.readthedocs.io/en/latest/getting_started.html.

Feel free to propose a PR.

acharles7:
@jimmylai Please close this one too.

PR #128, opened 2020-09-11
Add console entry points for pipx compatibility

gaborbernat:
Resolves https://github.com/Instagram/Fixit/issues/124

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,6 +46,12 @@ setuptools.setup(
             if "=" in dep
         ],
     },
+    entry_points={
+        "console_scripts": [
+            "fixit-run=fixit.cli::run_rules",
+            "fixit-apply=fixit.cli::apply_fix",
+        ],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -46,6 +46,12 @@
             if "=" in dep
         ],
     },
+    entry_points={
+        "console_scripts": [
+            "fixit-run=fixit.cli::run_rules",
+            "fixit-apply=fixit.cli::apply_fix",
```
jimmylai, line 52: Thanks for contributing.
This should be just one single colon.

Please provide a test plan and result in PR summary. 
You can do `python setup.py sdist` and then use pipx to install the built artifact to verify whether it works.

Please also sign the CLA.

gaborbernat:
Sorry signing the cla would be too troublesome for me, so closing.

PR #130, opened 2020-09-17
Relax install dependency requirements

jimmylai:
## Summary
Relax install dependency requirements. So it's easier to install.

## Test Plan

CI builds.


jimmylai:
This is experimental and no longer needed.

PR #140, opened 2020-09-26
Add NoStringTypeAnnotationRule

liangming168:
## Summary
Enforce the use of type identifier instead of using string type hints.

## Test Plan
unit test: tox -e py37

Screen Shot 2020-09-25 at 9 20 49 PM


liangming168:
__future__ feature annotations is not defined before python 3.7, any idea to handle this case?

fixit/rules/no_string_type_annotation.py
```diff
@@ -0,0 +1,303 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from __future__ import annotations
```
jimmylai, line 6: This line can be removed in order to support Python 3.6.

jimmylai (changes requested):
Please don't commit `.DS_Store` files. They are unrelated.

fixit/rules/no_string_type_annotation.py
```diff
@@ -0,0 +1,303 @@
...
+        if not self.has_future_annotations_import:
+            return
+        if original_node in self.in_literal:
+            self.in_literal.remove(original_node)
+
+    def visit_SimpleString(self, node: cst.SimpleString) -> None:
+        if not self.has_future_annotations_import:
+            return
+        if self.in_annotation and not self.in_literal:
+            # This is not allowed past Python3.7 since its no longer necessary.
```
jimmylai, line 296:
```suggestion
            # This is not allowed past Python3.7 since it's no longer necessary.
```

fixit/rules/no_string_type_annotation.py
```diff
@@ -0,0 +1,303 @@
...
+    CstContext,
+    CstLintRule,
+    InvalidTestCase as Invalid,
+    ValidTestCase as Valid,
+)
+
+
+class NoStringTypeAnnotationRule(CstLintRule):
+    """
+    Enforce the use of type identifier instead of using string type hints.
```
jimmylai, line 24:
```suggestion
    Enforce the use of type identifier instead of using string type hints for simplicity and better syntax highlighting.
    Starting in Python 3.7, `from __future__ import annotations` can postpone evaluation of type annotations  
    `PEP 563 <https://www.python.org/dev/peps/pep-0563/#forward-references>`_
    and thus forward references no longer need to use string annotation style.
```

jimmylai:
Thanks for contributing! Other than the comments, LGTM.

liangming168:
> Please don't commit `.DS_Store` files. They are unrelated.

`.DS_Store` files are all deleted.

fixit/rules/no_string_type_annotation.py
```diff
@@ -0,0 +1,304 @@
...
+    InvalidTestCase as Invalid,
+    ValidTestCase as Valid,
+)
+
+
+class NoStringTypeAnnotationRule(CstLintRule):
+    """
+    Enforce the use of type identifier instead of using string type hints for simplicity and better syntax highlighting.
+    Starting in Python 3.7, `from __future__ import annotations` can postpone evaluation of type annotations
+    `PEP 563 <https://www.python.org/dev/peps/pep-0563/#forward-references>`
```
jimmylai, line 24: Fix RST syntax. Note that the `_` is required for inline link.
https://docutils.sourceforge.io/docs/user/rst/quickref.html#hyperlink-targets
```suggestion
    Starting in Python 3.7, ``from __future__ import annotations`` can postpone evaluation of type annotations
    `PEP 563 <https://www.python.org/dev/peps/pep-0563/#forward-references>`_
```

jimmylai (approved):
LGTM

PR #143, opened 2020-10-03
Ensure first lines remain intact with AddMissingHeaderRule

thatch:
## Summary

Python files commonly start with something like `#!/usr/bin/env python` and this is only effective if on the very first line of the file.  There may be other directive comments like `# pyre-strict` that need to exist near the top as well; this does not yet handle that or allow configuration.

## Test Plan

Added a new test, failing on master.


jimmylai (approved):
Makes sense.
Thanks for contributing!

PR #144, opened 2020-10-07
Add UseAssertIsNotNoneRule

nt591:
## Summary

Adds a lint rule to check for the pattern `self.assertTrue(x is not None)` and corrects to `self.assertIsNotNone(x)`

## Test Plan

`tox -e py38`

```
----------------------------------------------------------------------
Ran 431 tests in 2.378s

OK
____________________________________________ summary ____________________________________________
  py38: commands succeeded
  congratulations :)
```

`tox -e py38 -- fixit.tests.UseAssertIsNotNoneRule`

```
----------------------------------------------------------------------
Ran 6 tests in 0.087s

OK
____________________________________________ summary ____________________________________________
  py38: commands succeeded
  congratulations :)
```

### DIFF
--- a/fixit/rules/use_assert_is_not_none.py
+++ b/fixit/rules/use_assert_is_not_none.py
@@ -65,10 +65,6 @@ class UseAssertIsNotNoneRule(CstLintRule):
     ]
 
     def visit_Call(self, node: cst.Call) -> None:
-        found_match = False
-        new_cst_attr = None
-        new_node_arg = None
-
         # `self.assertTrue(x is not None)` -> `self.assertIsNotNone(x)`
         if m.matches(
             node,
@@ -85,9 +81,14 @@ class UseAssertIsNotNoneRule(CstLintRule):
                 ],
             ),
         ):
-            found_match = True
-            new_node_arg = cst.Arg(ensure_type(node.args[0].value, cst.Comparison).left)
-            new_cst_attr = "assertIsNotNone"
+            new_call = node.with_changes(
+                func=cst.Attribute(
+                    value=cst.Name("self"), attr=cst.Name("assertIsNotNone")
+                ),
+                args=[cst.Arg(ensure_type(node.args[0].value, cst.Comparison).left)],
+            )
+            self.report(node, replacement=new_call)
+
         # `self.assertFalse(x is None)` -> `self.assertIsNotNone(x)`
         elif m.matches(
             node,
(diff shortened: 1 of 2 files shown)


jimmylai:
This new rule successfully find an existing violation and that's cause of the build failure.
Let's fix the existing code.
```
fixit/common/tests/test_imports.py:61:9
    UseAssertIsNotNoneRule: "assertTrue" is deprecated. Use "assertIsNotNone"
    instead.
    See https://docs.python.org/2/library/unittest.html#deprecated-aliases
```

fixit/rules/use_assert_is_not_none.py
```diff
@@ -0,0 +1,71 @@
...
+class UseAssertIsNotNoneRule(CstLintRule):
+    """
+    Discourages use of ``assertTrue(x is not None)`` as it is deprecated (https://docs.python.org/2/library/unittest.html#deprecated-aliases).
+    Use ``assertIsNotNone(x)`` instead.
+
+    """
+
+    MESSAGE: str = (
+        '"assertTrue" is deprecated. Use "assertIsNotNone" instead.\n'
+        + "See https://docs.python.org/2/library/unittest.html#deprecated-aliases"
```
jimmylai, line 23: Let's link to Python 3 doc both in this message and the docstring.
https://docs.python.org/3.8/library/unittest.html#deprecated-aliases

jimmylai (commented):
This looks great.
There is also an `assertIsNone` for replacing `assertFalse(a is not None)` (which is also deprecating) or `assertTrue(a is None)`.
It can be in the same rule or as a separate rule.
Make them in the same rule can make the future maintenance easier.

nt591:
There's a unique type error on Ubuntu 3.7 so I'm going to push a small refactor that increases duplication but removes Optional type inference

fixit/rules/use_assert_is_not_none.py
```diff
@@ -0,0 +1,161 @@
...
+    MESSAGE: str = (
+        '"assertTrue" and "assertFalse" are deprecated. Use "assertIsNotNone" and "assertIsNone" instead.\n'
+        + "See https://docs.python.org/3.8/library/unittest.html#deprecated-aliases"
+    )
+
+    VALID = [
+        Valid("self.assertTrue(not x is None)"),
+        Valid("self.assertTrue(x)"),
+        Valid("self.assertFalse(not x is None)"),
+        Valid("self.assertFalse(x)"),
```
jimmylai, line 30: `not x is None` is the same as `x is not None`
It can actually be also handled by this rule to provide an autofix.
Let's also add matcher pattern for it.

Use assertTrue/assertFalse as an example may cause confusion to users since they are deprecated.
(The Valid and Invalid cases are generated as examples in the documentation like this)
https://fixit.readthedocs.io/en/latest/rules/ComparePrimitivesByEqualRule.html
Let's use the suggested `assertIsNotNone` and `assertIsNone` as Valid case examples.

nt591: Going to add a few more ensure_types to see if I can get the typecheck to pass on Ubuntu + 3.7

jimmylai (approved):
Looks good.

One improvement can be made is to use `SaveMatchedNode` and `extract` to get the node in the matched tree to keep the code simpler and avoid using the complex attribute access with multiple `ensure_type`.
https://libcst.readthedocs.io/en/latest/matchers.html#libcst.matchers.SaveMatchedNode

PR #146, opened 2020-10-08
UseAssertIs{Not}None - use m.extract and m.SaveMatchedNode to memoize tree traversal

nt591:
## Summary

Given feedback from https://github.com/Instagram/Fixit/pull/144

I used the m.extract and m.SaveMatchedNode patterns. Unfortunately, accessing the children of those nodes via dot-accessor didn't work under Pyre, since m.SaveMatchedNode returns a Sequence of cst.BaseNode. This means:

1) we need to unwrap the Sequence
2) Since the Sequence is of the base type, we still need to use `ensure_type` to coerce the typechecker to play along

## Test Plan

```
ƛ No type errors found
```

`tox -e py38`

```
Ran 444 tests in 2.345s

OK
________________________________________________________________________ summary ________________________________________________________________________
  py38: commands succeeded
  congratulations :)
```

### DIFF
--- a/fixit/rules/use_assert_is_not_none.py
+++ b/fixit/rules/use_assert_is_not_none.py
@@ -3,6 +3,8 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+from typing import Sequence
+
 import libcst as cst
 import libcst.matchers as m
 from libcst.helpers import ensure_type
@@ -47,6 +49,10 @@ class UseAssertIsNotNoneRule(CstLintRule):
             expected_replacement="self.assertIsNotNone(f())",
         ),
         Invalid(
+            "self.assertTrue(not x is not None)",
+            expected_replacement="self.assertIsNone(x)",
+        ),
+        Invalid(
             "self.assertTrue(f(x) is not None)",
             expected_replacement="self.assertIsNotNone(f(x))",
         ),
@@ -66,6 +72,10 @@ class UseAssertIsNotNoneRule(CstLintRule):
             expected_replacement="self.assertIsNone(f())",
         ),
         Invalid(
+            "self.assertFalse(not x is not None)",
+            expected_replacement="self.assertIsNotNone(x)",
+        ),
+        Invalid(
             "self.assertFalse(f(x) is not None)",
             expected_replacement="self.assertIsNone(f(x))",
         ),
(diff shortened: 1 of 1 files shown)


fixit/rules/use_assert_is_not_none.py
```diff
@@ -76,173 +86,86 @@ class UseAssertIsNotNoneRule(CstLintRule):
...
+                comparison_type = comparison_type[0]
+
+            if m.matches(argument, m.Comparison()):
+                assertion_argument = ensure_type(argument, cst.Comparison).left
+            else:
+                assertion_argument = ensure_type(
+                    ensure_type(argument, cst.UnaryOperation).expression, cst.Comparison
+                ).left
 
+            negations_seen = 0
```
nt591, line 156: The linter didn't like my first attempt with the many nested branches however this ends up being a lot cleaner.

Every "negation" determines if the node should be `assertIsNone` vs `assertIsNotNone`. I can just increment negations, knowing that 

1) assertFalse adds a negation
2) a UnaryOperation (that is, `not x is None`) adds a negation
3) using `is not` adds a negation

fixit/rules/use_assert_is_not_none.py
```diff
@@ -76,173 +86,86 @@ class UseAssertIsNotNoneRule(CstLintRule):
...
-                func=cst.Attribute(
-                    value=cst.Name("self"), attr=cst.Name("assertIsNone")
-                ),
-                args=[cst.Arg(ensure_type(node.args[0].value, cst.Comparison).left)],
-            )
-            self.report(node, replacement=new_call)
+        if result:
+            assertion_name = result["assertion_name"]
+            if isinstance(assertion_name, Sequence):
+                assertion_name = assertion_name[0]
```
jimmylai, line 139: Based on the description of `extract`, the value in the result dict can be either a node or a sequence. So use isinstance to teach Pyre is the right thing to do.
https://libcst.readthedocs.io/en/latest/matchers.html#libcst.matchers.extract

fixit/rules/use_assert_is_not_none.py
jimmylai: This is really neat!

fixit/rules/use_assert_is_not_none.py
```diff
@@ -76,173 +86,86 @@ class UseAssertIsNotNoneRule(CstLintRule):
...
+                                        comparisons=[
+                                            m.ComparisonTarget(
+                                                m.SaveMatchedNode(
+                                                    m.OneOf(m.Is(), m.IsNot()),
+                                                    "comparison_type",
+                                                ),
+                                                comparator=m.Name("None"),
+                                            )
+                                        ]
+                                    ),
```
jimmylai, line 126: The entire `m.Comparison(...)` is the same as the one used above.
We can move it as a variable (e.g. `match_compare_is_none = m.Comparison(...)`) and use the variable in both places  to remove the duplication.

jimmylai (approved):
Looks great!
Found one place can be further improved.

PR #151, opened 2020-10-23
[config] add use_noqa and defaults to False

jimmylai:
## Summary
Add use_noqa and defaults to False

## Test Plan
`tox -e py38`


Kronuz (approved):
Nice!

PR #153, opened 2020-10-28
bump version to 0.1.2

zsol:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+# 0.1.2 - 2020-10-29
+
+## New Rules
+- Add SortedAttributesRule [#149](https://github.com/Instagram/Fixit/pull/149)
+
+## Added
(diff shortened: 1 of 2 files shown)


jimmylai (approved):
LGTM

jimmylai:
There is one more merged PR that is not mentioned in the change set and it refactored UseAssertIsNotNoneRule using matcher to make it simpler.
https://github.com/Instagram/Fixit/commit/8a11a0969d943557c75a18438432975d4b4911db

zsol:
I intentionally left that out because it seemed like an internal refactor that users wouldn't care much about. Do you think it should still be in the changelog?

jimmylai:
> I intentionally left that out because it seemed like an internal refactor that users wouldn't care much about. Do you think it should still be in the changelog?

Make sense. I don't have strong preference.

PR #154, opened 2020-11-03
Add NoRedundantArgumentsSuperRule

mkniewallner:
## Summary
Ensure that calls to `super` that use the direct parent as a first argument are reported (and autofixed when requested), according to [PEP 3135](https://www.python.org/dev/peps/pep-3135/).

This will turn
```python
class Foo(Bar):
    def foo(self, bar):
        super(Foo, self).foo(bar)
```
into
```python
class Foo(Bar):
    def foo(self, bar):
        super().foo(bar)
```

The rule also takes into account nested classes, so that this code
```python
class Foo:
    class InnerFoo(Bar):
        def foo(self, bar):
            super(Foo.InnerFoo, self).foo(bar)
```
gets turned into
```python
class Foo:
    class InnerFoo(Bar):
        def foo(self, bar):
            super().foo(bar)
```

## Test Plan
Test cases have been added to ensure that correctly formatted code is not modified and code that matches the rule reports the issue.

### DIFF
--- a/fixit/rules/no_redundant_arguments_super.py
+++ b/fixit/rules/no_redundant_arguments_super.py
@@ -1,0 +1,155 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import List, Union
+
+import libcst as cst
+import libcst.matchers as m
+
+from fixit import (
+    CstContext,
+    CstLintRule,
+    InvalidTestCase as Invalid,
+    ValidTestCase as Valid,
+)
+
+
+class NoRedundantArgumentsSuperRule(CstLintRule):
+    """
+    Remove redundant arguments when using super for readability.
+    """
+
+    MESSAGE: str = (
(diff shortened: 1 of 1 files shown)


jimmylai (approved):
Thanks for contributing!
The test cases are comprehensive and make sense.

This rule reminds me some other conventions that can be built as Fixit rules.
1. a method in a class should have `self` as the first argument.
2. a classmethod in a class should have `cls` as the first argument.
3. a staticmethod in a class should not have `self` nor `cls` as the first argument.

Feel free to build more rules for more Python conventions to help developers write better code faster.

mkniewallner:
> 2. a classmethod in a class should have `cls` as the first argument.

This one is actually [already implemented](https://github.com/Instagram/Fixit/blob/v0.1.2/fixit/rules/cls_in_classmethod.py) 🙂

jimmylai:
> > 1. a classmethod in a class should have `cls` as the first argument.
> 
> This one is actually [already implemented](https://github.com/Instagram/Fixit/blob/v0.1.2/fixit/rules/cls_in_classmethod.py) 🙂

That's great! I almost forget it.

PR #156, opened 2020-11-18
Allow glob patterns instead of parent dirs for matching configs

Kronuz:
## Summary

Allow using globs as configuration for rules, not only parent directories.

This allows for creating specific rules for specific files, not only directories. For example:

```
{
    "*/urls.py": {
        "rules": [
            ["*", "deny"]
        ]
    }
}
```

Will allow specifying rules for all `urls` modules anywhere.

---

This also allows specifying if the allow or deny rule are meant for global or local scopes only:

```
{
    "*/urls.py": {
        "rules": [
            ["common.*", "allow"],
            ["*", "deny global"]
        ],
        "message": "Module in {current_file} only allows certain global imports. {imported} cannot be imported here."
    }
}
```

Allowing to import `common.*` anywhere and everything else only inside functions, as inner imports.

## Test Plan

Run tests

### DIFF
--- a/fixit/common/testing.py
+++ b/fixit/common/testing.py
@@ -115,6 +115,13 @@ class LintRuleTestCase(unittest.TestCase):
                 raise AssertionError(
                     f"Expected:\n    {test_case.expected_str}\nBut found:\n    {report}"
                 )
+            if (
+                test_case.expected_message is not None
+                and test_case.expected_message != report.message
+            ):
+                raise AssertionError(
+                    f"Expected message:\n    {test_case.expected_message}\nBut got:\n    {report.message}"
+                )
 
             validate_patch(report, test_case)
 
--- a/fixit/rules/import_constraints.py
+++ b/fixit/rules/import_constraints.py
@@ -5,6 +5,7 @@
 
 import os
 from dataclasses import dataclass
+from enum import Enum
 from fnmatch import fnmatch
 from functools import lru_cache
 from pathlib import Path
@@ -12,6 +13,7 @@ from typing import Dict, Iterable, List, Optional, Sequence, Set
 
 import libcst as cst
 from libcst.helpers import get_full_name_for_node_or_raise
+from libcst.metadata import GlobalScope, ScopeProvider
 
 from fixit import (
     CstContext,
@@ -25,6 +27,15 @@ from fixit import (
 TEST_REPO_ROOT: str = str(Path(__file__).parent.parent)
 
 
+class RuleAction(Enum):
+    ALLOW = "allow"  # allow everywhere
+    DENY = "deny"  # deny everywhere
+    ALLOW_GLOBAL = "allow_global"  # allow in global scope
+    ALLOW_LOCAL = "allow_local"  # allow in local scopes
+    DENY_GLOBAL = "deny_global"  # deny in global scope
+    DENY_LOCAL = "deny_local"  # deny in local scopes
+
+
 def _gen_testcase_config(dir_rules: Dict[str, object]) -> LintConfig:
     return LintConfig(
         repo_root=TEST_REPO_ROOT, rule_config={"ImportConstraintsRule": dir_rules}
@@ -34,7 +45,7 @@ def _gen_testcase_config(dir_rules: Dict[str, object]) -> LintConfig:
 @dataclass(frozen=True)
 class _ImportRule:
     pattern: str  # dot-separated module/package name prefix, or wildcard (*)
-    allow: bool
+    action: RuleAction
 
     @staticmethod
     def from_config(rule: object) -> "_ImportRule":
@@ -46,9 +57,7 @@ class _ImportRule:
             )
         else:
             module, action = rule
-            if action not in ("allow", "deny"):
-                raise ValueError("A rule should either allow or deny a pattern")
-            return _ImportRule(module, action == "allow")
+            return _ImportRule(module, RuleAction(action))
 
     @property
     def is_wildcard(self) -> bool:
@@ -67,6 +76,7 @@ class _ImportConfig:
     rules: Sequence[_ImportRule]
     ignore_tests: bool
     ignore_types: bool
+    message: Optional[str]
 
     @staticmethod
     def from_config(settings_for_dir: Dict[object, object]) -> "_ImportConfig":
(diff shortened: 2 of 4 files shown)


fixit/rules/import_constraints.py
```diff
@@ -328,13 +334,18 @@ def __init__(self, context: CstContext) -> None:
...
+            # found in the config. The closest match in the settings will be used.
+            parts = list(self._abs_file_path.parts)
+            while parts:
+                part_dir = os.path.join(*parts)
+                parts.pop()
+                for pattern, settings_for_dir in formatted_config.items():
+                    if fnmatch(part_dir, pattern):
+                        self._config = _ImportConfig.from_config(settings_for_dir)
+                        parts.clear()
+                        break
```
jimmylai, line 348: When a match is found, we should also break the `while parts` loop.

jimmylai (commented):
There are lint errors.

fixit/rules/import_constraints.py
Kronuz: `parts.clear()` does that.

Kronuz: These are the times I wish Python had `break 2`, like PHP does, or `break <label>` like Swift. I can also use the `else: continue; break` idiom if it’s clearer.

fixit/rules/import_constraints.py
```diff
@@ -328,13 +335,20 @@ def __init__(self, context: CstContext) -> None:
...
+            # given configurations' glob patterns, stopping early if a match is
+            # found in the config. The closest match in the settings will be used.
+            parts = list(self._abs_file_path.parts)
+            while parts:
+                part_dir = os.path.join(*parts)
+                parts.pop()
+                for pattern, settings_for_dir in formatted_config.items():
+                    if fnmatch(part_dir, pattern):
+                        self._config = _ImportConfig.from_config(settings_for_dir)
+                        break
```
jimmylai, line 348: Even though the else continue break works, it doesn't look strait-forward to me. How about just check whether we already find a config for self._config? That looks easier to understand to me.

```suggestion
            while parts:
                part_dir = os.path.join(*parts)
                parts.pop()
                for pattern, settings_for_dir in formatted_config.items():
                    if fnmatch(part_dir, pattern):
                        self._config = _ImportConfig.from_config(settings_for_dir)
                        break
                if self._config is not None:
                    break
```

jimmylai:
The CI builds has a test case failure.
https://github.com/Instagram/Fixit/pull/156/checks?check_run_id=1419755011

It looks like a bug in the new version of LibCST related to the new Assignment/Access changes. CC @zsol 

It changes `b = [a for a in [1,2,3]]` as `b = [cls for a in [1,2,3]]` which is wrong. The local access of `a` in the comprehension should refer to the local definition of var `a` in the comprehension.

jimmylai:
@Kronuz there is a type check error regarding `fnmatch`.
https://github.com/Instagram/Fixit/pull/156/checks?check_run_id=1483019512

fixit/rules/import_constraints.py
```diff
@@ -314,6 +390,56 @@ class ImportConstraintsRule(CstLintRule):
...
+        # Custom message
+        Invalid(
+            """
+            import common
+            """,
+            config=_gen_testcase_config(
+                {
+                    "dir": {
+                        "rules": [["*", "deny"]],
+                        "message": "'{imported}' cannot be imported from '{current_file}'",
```
jimmylai, line 436: I like the idea to allow custom message.
Can you update the docstring of ImportConstraintsRule and explain how to use the message?
We should explain that `imported` and `current_file` are two variables that can be embedded in the message.

fixit/common/testing.py
```diff
@@ -116,6 +116,11 @@ def _test_method(
                     f"Expected:\n    {test_case.expected_str}\nBut found:\n    {report}"
                 )
 
+            if test_case.message is not None and test_case.message != report.message:
+                raise AssertionError(
+                    f"Expected message:\n    {test_case.message}\nBut got:\n    {report.message}"
+                )
+
```
jimmylai, line 123: The message in InvalidTestCase is only used in one of the new test case. Do we really need to add this new field?
Ideally, the test case example code should explain itself clearly. The example code can also have some comments to explain it better. I'd suggest not to add InvalidTestCase.message to keep it simpler.

fixit/rules/import_constraints.py
```diff
@@ -24,6 +27,30 @@
...
+    "allow global": Action.ALLOW_GLOBAL,
+    "allow local": Action.ALLOW_LOCAL,
+    "deny global": Action.DENY_GLOBAL,
+    "deny local": Action.DENY_LOCAL,
+    # Aliases:
+    "allow globally": Action.ALLOW_GLOBAL,
+    "allow locally": Action.ALLOW_LOCAL,
+    "deny globally": Action.DENY_GLOBAL,
+    "deny locally": Action.DENY_LOCAL,
+}
```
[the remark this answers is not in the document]
Kronuz: Yes, the idea behind the map was to also allow other aliases (such as "allow globally", "allow global", and "allow_global"). But yes, we can use the simple Enum, if we don't want that support.

fixit/common/testing.py
Kronuz: I wanted to test the custom messages are working, not only show they exist. This allows testing custom messages and it doesn't add much complexity, only a new field that can be tested. we could call the field "`expected_message`" though. What do you think?

fixit/rules/import_constraints.py
```diff
@@ -45,9 +72,11 @@ def from_config(rule: object) -> "_ImportRule":
...
         else:
             module, action = rule
-            if action not in ("allow", "deny"):
-                raise ValueError("A rule should either allow or deny a pattern")
-            return _ImportRule(module, action == "allow")
+            if action not in ACTIONS_MAP:
+                raise ValueError(
+                    "A rule should either allow or deny a pattern (possibly globally or locally)"
+                )
+            return _ImportRule(module, ACTIONS_MAP[action])
```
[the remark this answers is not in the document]
Kronuz: We could make this whole block even simpler if we don't want that map for aliases, albeit with a more generic error (`ValueError: 'invalid_action' is not a valid RuleAction`):

```
            return _ImportRule(module, Action(action))
```

fixit/common/testing.py
jimmylai: Yeah, `expected_message` is a better name. I can see it can be useful for rules provide custom messages.

fixit/rules/import_constraints.py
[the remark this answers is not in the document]
jimmylai: Not supporting different alias doesn't seem to be a big issue. It can also keep the config more consistent.

jimmylai (approved):
LGTM
When releasing this, we need to update existing config file to avoid the new RuleAction value error.

Kronuz:
> LGTM
> When releasing this, we need to update existing config file to avoid the new RuleAction value error.

What do you mean?

jimmylai:
> > LGTM
> > When releasing this, we need to update existing config file to avoid the new RuleAction value error.
> 
> What do you mean?

I read it wrong, so there won't be value error since we keep the lower case allow/deny. We can review existing rule to see if we need to refine them.

PR #161, opened 2020-12-07
add UseLintFixmeCommentRule

jimmylai:
## Summary
add UseLintFixmeCommentRule to warn on deprecated noqa comment.
Converting a noqa to lint-fixme requires extra information (lint rule name) which makes it hard to provide autofix.
It also requires moving inline noqa comments to a standalone line.
We could consider provide a VSCode plugin as follow up that can suggest adding a lint-fixme or lint-ignore given a lint warning as a second option.

## Test Plan
`tox -e py38 -- fixit.tests.UseLintFixmeCommentRule`

### DIFF
--- a/fixit/rules/use_lint_fixme_comment.py
+++ b/fixit/rules/use_lint_fixme_comment.py
@@ -1,0 +1,61 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import libcst as cst
+
+from fixit import CstLintRule, InvalidTestCase as Invalid, ValidTestCase as Valid
+
+
+class UseLintFixmeCommentRule(CstLintRule):
+    """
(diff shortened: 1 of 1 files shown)


Kronuz (approved):
I guess you could use that regex instead, to handle when users use `#noqa` (no space, or multiple spaces) or when users put noqa after another comment. But for simplicity, I think handling uppercase/lowercase `# noqa` could be enough.

jimmylai:
> I guess you could use that regex instead, to handle when users use `#noqa` (no space, or multiple spaces) or when users put noqa after another comment. But for simplicity, I think handling uppercase/lowercase `# noqa` could be enough.
 
The regex requires one specific space before noqa. For the case of noqa after another comment, as long as the comment starts with noqa, it's a valid noqa comment which we want to show this warning.
https://github.com/Instagram/Fixit/blob/ef5edca7c6d3cc5f2f1cb5ff54b769ac8bb0d923/fixit/common/config.py#L34

PR #164, opened 2020-12-15
Draft for IG Lint FYI workplace discussion

mattfielder:
## Summary
Trying to supporting having lint autofix rules support multiple reports, so for instance you could change a variable named during assignment and have any subsequent access also updated, talking about this with Jimmy in workplace this is just a draft for discussion.


lisroach:
I think this is pretty old now, can we close this PR?

PR #172, opened 2021-02-02
refactor: search yaml using `pathlib.Path.parents`

pwoolvett:


Kronuz:
This looks awesome! and much cleaner.

PR #175, opened 2021-02-25
Expose `ipc_main` with arguments instead of argparse

aehernandez:
## Summary

Refactored `ipc_main` into `run_ipc` which takes in arguments instead of relying on argparse. `ipc_main` still exists for backwards compatibility. 

## Test Plan

Added a test verifying that
1. The new function runs on a single path and returns output as expected
2. The new function still supports reporting on multiple input paths

Ran `tox -e py38` to verify tests were running and green.

### DIFF
--- a/fixit/cli/__init__.py
+++ b/fixit/cli/__init__.py
@@ -199,7 +199,16 @@ class IPCResult:
     paths: List[str]
 
 
-def ipc(opts: LintOpts, paths: List[str], prefix: str = None, workers: LintWorkers) -> IPCResult:
+def run_ipc(opts: LintOpts, paths: List[str], prefix: str = None, workers: LintWorkers) -> IPCResult:
+    """
+    Given a LintOpts config with lint rules and lint success/failure report formatter,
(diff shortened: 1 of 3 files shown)


fixit/cli/__init__.py
```diff
@@ -239,4 +237,26 @@ def ipc_main(opts: LintOpts) -> IPCResult:
...
+        fromfile_prefix_chars="@",
+        parents=[get_multiprocessing_parser()],
+    )
+    parser.add_argument("paths", nargs="*", help="List of paths to run lint rules on.")
+    parser.add_argument("--prefix", help="A prefix to be added to all paths.")
+    args: argparse.Namespace = parser.parse_args()
+
+    return run_ipc(
+        opts=opts, paths=args.paths, prefix=args.prefix, workers=args.workers
+    )
```
jimmylai, line 262: Let's deprecate this.

aehernandez: 👍 Added a deprecation warning using python's standard `warnings.warn` functionality.

mvismonte (approved):
Looks good!

Issue #125, opened 2020-09-10
Incorrect rewriting of dict(test=[])


wolfgangl:
Example interaction: 
```
echo "dict(test=[])" > test.py
python -m  fixit.cli.apply_fix test.py 
Scanning 1 files
test.py

Encountered exception <class 'Exception'> for the following paths:
test.py
Running `pyre start` may solve the issue.
test.py:1:1 [applied fix]
    RewriteToLiteralRule: It's unnecessary to use a list or tuple within a call to dict sincethere is
    literal syntax for this type
Encountered exception <class 'Exception'> for the following paths:
test.py
test.py
test.py
Running `pyre start` may solve the issue.
All done! ✨ 🍰 ✨
1 file left unchanged.

Found 1 reports in 1 files in 0.33 seconds.
(py3_venv) wolf@ldp-wolf:~/ros_ws_py3/src/real_world$ cat test.py
{}
(py3_venv) wolf@ldp-wolf:~/ros_ws_py3/src/real_world$ 
```

RewriteToLiteralRule is losing a field in the dictionary.

jimmylai:
Nice catch!
The current implementation didn't consider the keyward argument carefully for the case like the example.
https://github.com/Instagram/Fixit/blob/f189e053d0b52f63231de9146bb6ca51984210a0/fixit/rules/rewrite_to_literal.py#L72-L76

A quick mitigation to avoid autofix the example case is to update the matcher pattern to require the `Arg` to to have no keyword name.
https://libcst.readthedocs.io/en/latest/nodes.html#libcst.Arg

The case `dict(test=[], other=1)` should be autofixed as `{"test": [], "other": 1}`. We can add some new logic handle this separately since it require convert the parameter name as SimpleString and it's specific to `dict()`.

Feel free to submit PRs to mitigate it and/or fix the autofixer.

amyreese:
Hey folks, we just released Fixit 2.0, which includes major changes to the core frameworks, linting engine, API, and CLI. In light of this new release, this issue is likely to be resolved or no longer relevant. Please take a look at the new [user guide](https://fixit.readthedocs.io/en/latest/guide.html#) or the [upgrade guide](https://fixit.readthedocs.io/en/latest/upgrade.html) for more information on how to migrate your projects to the new version.

If you believe this issue is still relevant or still occurring in the latest version, please feel free to open a new issue, or reply here with more details.

Thank you!

PR #176, opened 2021-03-15
Adds support for keyword arguments in dicts.

lisroach:
## Summary
Fix for issue #125.

My first time working with Fixit/LibCST so I appreciate any feedback on how this could be improved.

Basic changes are to check if there are any keyword arguments in the `dict()` call and swap it to `{"keyword": "blah"}`.

## Test Plan
Added test cases, tested with:
`ox -e py38 -- fixit.tests.RewriteToLiteralRule`

and ran

`tox -e autofix`

Also running (from the issue):
```
$ echo "dict(test=[])" > test.py
$ python -m  fixit.cli.apply_fix test.py
```
now works.

### DIFF
--- a/fixit/rules/rewrite_to_literal.py
+++ b/fixit/rules/rewrite_to_literal.py
@@ -13,7 +13,7 @@ UNNECESSARY_LITERAL: str = (
     "It's unnecessary to use a list or tuple within a call to {func} since"
     + " there is literal syntax for this type"
 )
-UNNCESSARY_CALL: str = (
+UNNECESSARY_CALL: str = (
     "It's slower to call {func}() than using the empty literal, because"
     + " the name {func} must be looked up in the global scope in case it has"
     + " been rebound."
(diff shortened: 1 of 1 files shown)


fixit/rules/rewrite_to_literal.py
```diff
@@ -64,14 +64,30 @@ class RewriteToLiteralRule(CstLintRule):
...
+            "dict(test=[], other=())", expected_replacement='{"test": [], "other": ()}'
+        ),
+        Invalid(
+            "dict(test=[], other=1)", expected_replacement='{"test": [], "other": 1}'
+        ),
+        Invalid(
+            "dict(other=1, test=[1, 2, 3])",
+            expected_replacement='{"other": 1, "test": [1, 2, 3]}',
+        ),
+        Invalid("dict([], other=1)", expected_replacement='{"other": 1}'),
```
jimmylai, line 78: Let's make this test case more useful via combining keys from the list and kwargs.
```suggestion
        Invalid("dict([('a', 1), ('b', 2)], other=1)", expected_replacement='{"a": 1, "b": 2, "other": 1}'),
```

fixit/rules/rewrite_to_literal.py
```diff
@@ -115,27 +146,44 @@ def visit_Call(self, node: cst.Call) -> None:
...
+                                lambda val: cst.DictElement(
+                                    val.elements[0].value, val.elements[1].value
+                                )
+                            )(
+                                cst.ensure_type(
+                                    ele.value,
+                                    cst.Tuple
+                                    if isinstance(ele.value, cst.Tuple)
+                                    else cst.List,
+                                )
```
jimmylai, line 182: Though this is existing code but it looks weird that `ensure_type` is used with `isinstance`.
`ensure_type` is for verifying the type and helps the type checker understanding the type. https://libcst.readthedocs.io/en/latest/helpers.html?highlight=ensure_type#libcst.helpers.ensure_type
The large `elif` branch can be separated as multiple smaller elif to handle different cases separately. That way, we can use different matcher of Tuple and List metches in different elif and remove the isinstance. It can also simplify the `if keyword_dict` to check it only once.
```
elif (
                len(elements) == 0
                or m.matches(
                    exp.args[0].value,
                    m.Tuple(elements=[pairs_matcher])
                    | m.List(elements=[pairs_matcher]),
                )
                or keyword_dict
            ):
```

jimmylai (changes requested):
There is a type error need to be fixed.
https://github.com/Instagram/Fixit/pull/176/checks?check_run_id=2116657092

Issue #181, opened 2021-04-05
Passing no providers to FullRepoMetadataConfig causes exception


lisroach:
My ultimate goal is to use Fixit without Pyre, and without having to pass `timeout=0` to the TypeInferenceProvider to force a breakage in the pyre subprocess call.

Created a simple repro:
```
from pathlib import Path
import logging
import sys
import libcst as cst
from fixit._version import FIXIT_VERSION
from fixit.common.config import get_lint_config
from fixit.common.config import get_rules_from_config
from fixit.common.full_repo_metadata import FullRepoMetadataConfig, get_repo_caches
from fixit.rule_lint_engine import lint_file
from libcst.metadata import MetadataWrapper, TypeInferenceProvider


def main(filename):
    full_repo_metadata_config = FullRepoMetadataConfig(
        providers=set(),  # <--- passing in no provider here instead of TypeInferenceProvider
        timeout_seconds=120,  # <--- If I were passing TypeInferenceProvider I would set this to 0
        logger=logging.getLogger("Metadata"),
    )

    with open(filename, "rb") as f:
        source = f.read()
    metadata_caches = get_repo_caches([filename], full_repo_metadata_config)
    try:
        cst_wrapper = MetadataWrapper(
            cst.parse_module(source),
            True,
            metadata_caches[filename],
        )
    except Exception:
        cst_wrapper = None
    rules = get_rules_from_config()
    config = get_lint_config()
    results = lint_file(
        Path(filename),
        source,
        rules=rules,
        config=config,
        cst_wrapper=cst_wrapper,
        find_unused_suppressions=True,
    )
    print(results)

if __name__ == "__main__":
    filename = sys.argv[0]
    main(filename)
```

Running this file results in:
```
(fixit-env) lisroach@lisroach-mbp Fixit % python3 hey.py hey.py
Traceback (most recent call last):
  File "hey.py", line 45, in <module>
    main(filename)
  File "hey.py", line 39, in main
    find_unused_suppressions=True,
  File "/Users/lisroach/Fixit/fixit/rule_lint_engine.py", line 115, in lint_file
    _visit_cst_rules_with_context(cst_wrapper, cst_rules, cst_context)
  File "/Users/lisroach/Fixit/fixit/rule_lint_engine.py", line 50, in _visit_cst_rules_with_context
    rule_instances, before_visit=before_visit, after_leave=after_leave
[...]

PR #182, opened 2021-04-05
Supports passing no providers to FullRepoMetadataConfig.

lisroach:
## Summary
https://github.com/Instagram/Fixit/issues/181

Allowing passing no providers to FullRepoMetadata. There is already a fallback if the provider is broken where a placeholder cache is used, so I used the same codepath to treat empty providers and exceptions the same way.

There is currently only one placeholder cache type so I hardcoded TypeInferenceProvider, but if we want this to be more flexible I can try to build support for multiple.

## Test Plan

Added a unit test. Tested manually and exception no longer fires and linters work correctly.

### DIFF
--- a/fixit/common/full_repo_metadata.py
+++ b/fixit/common/full_repo_metadata.py
@@ -53,7 +53,7 @@ def get_repo_caches(
     caches = {}
     paths_iter = iter(paths)
     head: Optional[str] = next(paths_iter, None)
-    placeholder: bool = False if config.providers else True
+    placeholder: bool = not config.providers
     while head is not None:
         paths_batch = tuple(chain([head], islice(paths_iter, config.batch_size - 1)))
         head = next(paths_iter, None)
--- a/fixit/common/full_repo_metadata.py
+++ b/fixit/common/full_repo_metadata.py
@@ -53,6 +53,7 @@ def get_repo_caches(
     caches = {}
     paths_iter = iter(paths)
     head: Optional[str] = next(paths_iter, None)
+    placeholder: bool = False if config.providers else True
     while head is not None:
         paths_batch = tuple(chain([head], islice(paths_iter, config.batch_size - 1)))
         head = next(paths_iter, None)
@@ -65,6 +66,7 @@ def get_repo_caches(
         try:
             frm.resolve_cache()
         except Exception:
+            placeholder = True
             # We want to fail silently since some metadata providers can be flaky. If a logger is provided by the caller, we'll add a log here.
             logger = config.logger
             if logger is not None:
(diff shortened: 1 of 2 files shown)


fixit/common/full_repo_metadata.py
```diff
@@ -53,6 +53,7 @@ def get_repo_caches(
     caches = {}
     paths_iter = iter(paths)
     head: Optional[str] = next(paths_iter, None)
+    placeholder: bool = False if config.providers else True
```
thatch, line 56: `not config.providers` maybe?

thatch:
Seems reasonable; check pyre and is there any other kind of full repo metadata other than TypeInferenceProvider that we might need to support?

lisroach:
I don't know why pyre is failing, I cannot get it to repro locally :(

I did some digging to try to figure out what the plan was with the `PLACEHOLDER_CACHES` and looks like it exists mostly because there is a bug in `TypeInferenceProviders` that triggers if cache is empty: https://github.com/Instagram/Fixit/pull/31#discussion_r452948963

I'm going to take a look at fixing that and reconfiguring this to not use placeholders at all if possible.

lisroach:
Created https://github.com/Instagram/LibCST/issues/475 for the LibCST bug

jimmylai:
Pyre failed due to a type error was found in fixit/rules/cls_in_classmethod.py:286:26 and that's related to the new LibCST release.
https://github.com/Instagram/Fixit/runs/2271833508
The same issue happens on master branch and it's better to fix it before merging new PRs.
FlattenSentinel is added in LibCST 0.3.18 and we need to update the type annotation of report().
https://github.com/Instagram/Fixit/blob/1d1cfe938fbc6f348a7666504982ec8ebd46cc87/fixit/common/base.py#L120

lisroach:
Thanks @jimmylai! No rush, I want to wait on this PR until after https://github.com/Instagram/LibCST/pull/476 is merged and released. I think it will clean up the code here nicely.

lisroach:
@zsol do you know when LibCST will be released again? I'd like to get this fix in :) Happy to help if I can!

lisroach:
There has been a new LibCST release so I am back looking at this. The bug in `TypeInferenceProviders` has been fixed but there is still a problem here: https://github.com/Instagram/LibCST/blob/e759ca8290585cf3badcd22ac5e1fdbc5e4f9e13/libcst/metadata/base_provider.py#L69

When there is a TimeoutError from using Pyre we trigger this exception that says `Cache is required for initializing TypeInferenceProvider.`

amyreese:

Issue #183, opened 2021-04-08
Discussion: Fixit Config File Requirements


lisroach:
When scaling Fixit to large monolithic repos we have run into challenges with the config file. In monorepos it is important to be able to do directory level configurations, that way users can tweak rules for just their projects without affecting the entire repository. 

I wanted to kick off a discussion around this problem first by getting an agreement on the requirements for fixit’s configs. Here is the problem space I wish to address:

*Problem: I want to turn on (or off) a lint rule just for my directory*

In Fixit today there is only one top-level config, and only one way to turn off rules for the repo: block_list_rules. 
This means we can turn off rules for everyone or no one, with no subdirectory level support. 

*Problem: I want to customize a rule_config  just for my directory*

Same as the issue above, since there is only one root-level config file there is no way to customize rules for subdirectories. 

# Proposals

## Inheritance of config files

The problems listed can be mostly addressed by supporting subdirectories with overriding configs. This way a top-level config file can exist that contains the defaults the majority of the repo will utilize, but subdirectories can override these defaults with their configurations.

Based on our experiences with Flake8, I propose these files support inheritance. When there is no config inheritance, users usually make copies of the top-level config to keep most of the defaults and only change the few they need. Unfortunately, these copies are not linked with the top-level config and get out of sync very quickly. Worse still, many users assume inheritance is a property already of the config files and do not copy the top-level configs at all, losing valuable and sometimes necessary customizations. 

Enabling inheritance of config files will enable custom configurations per-directory, while keeping in sync with the top-level config and its defaults.

Inheritance opens up a lot of questions, which I have listed in the next section below.

## Create an `allow_list_rules` key

To fully address the first problem I propose we add a new key, `allow_list_rules`.
Adding an `allow_list_rules` section to the config will enable us to mark rules as ‘on’, instead of defaulting to everything running. Using this, directories can turn on rules they want without the rule also being turned on everywhere else.

This will also make it less risky to upgrade to a new version of Fixit. New rules can be added to Fixit, but they won’t start firing until we explicitly turn them on. This gives us a chance to test and vet new rules. 
 
`block_list_rules` should take precedence. This way inherited rules that are on can be overridden and turned off, as well as it enables the root config to turn off dangerous rules for all leaf configs in a single location.

# Inheritance Questions

Assuming people are on board with the idea of inheritance, there are several questions that should be discussed.

### Should leaf configs override the root, or merge with the root config?

Example:
`root/.fixit.config.yaml`

```json
[...]

PR #184, opened 2021-04-21
Adds allow_list_rules setting for configs.

lisroach:
## Summary
Adds a new setting `allow_list_rules` that enables setting specific rules as "on". This solves a couple things:
- Safer upgrades of Fixit versions for users because newly added rules can be turned off by default
- Along with inheritable configs (to be implemented https://github.com/Instagram/Fixit/issues/183) this gives finer grain control over what rules run in individual directories

If `allow_list_rules` is left out of a config or set to `[]` it defaults to allow all to keep backwards compatibility.

## Test Plan
Added new unit tests. Also updated the docs.


lisroach:
`tox -e py38`
`tox -e py37`
`tox -e autofix`

All pass locally. Are the issues unrelated here?

thatch:
@lisroach the windows failures are because of the multiline strings with escaped newlines.  It's using the git mode that checks it out with dos newlines, so instead of `<backslash><LF>` it is `<backslash><CR><LF>`, and it's complaining that `<backslash><CR>` is an invalid escape.  I verified that it does not do the right thing when run, either.  (It seems to omit the CR, but leave the subsequent LF.)

I think elsewhere in fixit and libcst they use textwrap.dedent and just use an indented multiline string.  I think blank lines are allowed in yaml, so could just omit the backslash too.

lisroach:
Thanks @thatch! That helped a lot.

The Pyre error is unrelated, from the latest LibCST upgrade based on Jimmy's comments from https://github.com/Instagram/Fixit/pull/182.  I'll work on a PR to fix that bug.

lisroach:
https://github.com/Instagram/Fixit/pull/187 for the Pyre error

lisroach:
https://github.com/Instagram/Fixit/pull/190 for the autofix errors.

docs/source/getting_started.ipynb
```diff
@@ -155,6 +155,10 @@
     "\n",
     "2. Next, you may wish to edit or add some specific settings. The available configurations are:\n",
     "\n",
+    "- ``allow_list_rules``: A list of rules (whether custom of from Fixit) that should be applied to the repository. Omitting this setting allows all rules to run. For example:\n",
+    "\n",
+    "    allow_list_rules: [Flake8PseudoLintRule]\n",
+    "\n",
     "- ``block_list_patterns``: A list of patterns that indicate that a file should not be linted. For example::\n",
     "\n",
     "    block_list_patterns: ['@generated', '@nolint']\n",
```
jimmylai, line 164: Can you also describe what will happen if a rule is added to both allow_list_rules and block_list_rules? That can help users understand the precedence order.

williamlw999-fb:
I'm not sure if block_list_rules should always take precedence when it comes to inheritance. Say a new lint is being rolled out that has a high impact area. A team may want to turn it off by default for their directory but enable it gradually in subdirectories. I think that as part of the inheritance, it may make sense to override the block / allow list in tandem. That is, each key / rule should only exist in at most one of those lists. Thus, when a sub-directory's config allow lists a previously block listed rule, the inheritance should recognize and remove the pre-existing block on the rule.

lisroach:
@williamlw999-fb interesting point! My main reasoning behind the ability for `block_list_rules` to take precedence is in the case of a dangerous rule. If we discover a rule is dangerous it can be turned off for an entire directory in one swoop, vs having to track down all the `allow_list_rules` that might contain it and turn it off individually.

One thing I think you might be misunderstanding is that a rule is not on by default if it does not exist in the `block_list_rules`. So if a parent config does not need to add a new rule to the `block_list_rules` in order to turn it off by default _as long as_ they have other rules in the `allow_list_rules`.

So if a parent config looks like:
```
block_list_rules: []
allow_list_rules: [MyRule1]
```
and a new rule named YourRule comes out, YourRule would _not_ be on for the subdirs. It is only on if it is added to `allow_list_rules`. So slowly rolling out a rule to subdirs should be as simple as adding to the subdirs' `allow_list_rules`.

The one thing I don't like about this is if you have:
```
block_list_rules: [MyRule1]
allow_list_rules: []
```
and you roll out YourRule, in this case YourRule is on everywhere. I would like to require `allow_list_rules` to always specify all the rules that are on, but in order to keep backwards compatibility we need to support `allow_list_rules` empty defaulting to everything on.

lisroach:
I'm going to merge this PR to unblock myself, I am happy to have more discussion on this and we can make changes before the next FixIt release if the current way isn't working.

Issue #186, opened 2021-04-22
Pyre Failing on master branch


lisroach:
Pyre failed due to a type error was found in fixit/rules/cls_in_classmethod.py:286:26 and that's related to the new LibCST release.
https://github.com/Instagram/Fixit/runs/2271833508
The same issue happens on master branch and it's better to fix it before merging new PRs.
FlattenSentinel is added in LibCST 0.3.18 and we need to update the type annotation of report().
https://github.com/Instagram/Fixit/blob/1d1cfe938fbc6f348a7666504982ec8ebd46cc87/fixit/common/base.py#L120

_Originally posted by @jimmylai in https://github.com/Instagram/Fixit/issues/182#issuecomment-814246741_

lisroach:
Fixed with #187

PR #187, opened 2021-04-22
Sentinel type error

lisroach:
## Summary
Pyre is broken on master due to changes in the latest version of LibCST. This fixes by adding `cst.FlattenSentinel` to some function calls.

https://github.com/Instagram/Fixit/issues/186

## Test Plan

### DIFF
--- a/fixit/common/autofix.py
+++ b/fixit/common/autofix.py
@@ -174,9 +174,7 @@ def _replace_or_remove(
     original_node: cst.CSTNode,
     replacement_node: Union[cst.CSTNode, cst.FlattenSentinel, cst.RemovalSentinel],
 ) -> cst.CSTNode:
-    if isinstance(replacement_node, cst.RemovalSentinel) or isinstance(
-        replacement_node, cst.FlattenSentinel
-    ):
+    if isinstance(replacement_node, (cst.RemovalSentinel, cst.FlattenSentinel)):
         return cst.ensure_type(parent.deep_remove(original_node), cst.CSTNode)
     else:
         return parent.deep_replace(original_node, replacement_node)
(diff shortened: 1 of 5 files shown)


thatch:
Consider updating requirements.txt to also depend on libcst>=0.3.18, where this name was added.

fixit/common/autofix.py
```diff
@@ -172,9 +172,11 @@ def _minimize_tail(self) -> "LintPatch":
 def _replace_or_remove(
     parent: cst.CSTNode,
     original_node: cst.CSTNode,
-    replacement_node: Union[cst.CSTNode, cst.RemovalSentinel],
+    replacement_node: Union[cst.CSTNode, cst.FlattenSentinel, cst.RemovalSentinel],
 ) -> cst.CSTNode:
-    if isinstance(replacement_node, cst.RemovalSentinel):
+    if isinstance(replacement_node, cst.RemovalSentinel) or isinstance(
+        replacement_node, cst.FlattenSentinel
+    ):
```
jimmylai, line 179: `tox -e autofix` should fix the lint warning on this line.

jimmylai: Can you add `python -m fixit.cli.apply_fix {posargs:.}` to the autofix target and then run it?
https://github.com/Instagram/Fixit/blob/master/tox.ini#L28-L34

That can make future development easier.

lisroach: `tox -e autofix` actually does not fix this line, which is strange. I'll update it by hand and add it to the `tox.ini` though.

jimmylai (approved):
Thanks for fixing!

PR #190, opened 2021-04-27
Autofixes three module doscstrings.

lisroach:
## Summary
For some reason `autofix` want to remove this trailing space and it is causing breakages on all of my PRs. Updating the space in its own separate PR here.
## Test Plan
`tox -e autofix`


jimmylai (approved):
The suggestions are probably from Black. LGTM.

Issue #191, opened 2021-04-30
`use_noqa` is never read from the config file


lisroach:
`use_noqa` was added in https://github.com/Instagram/Fixit/pull/151 but when the config file is read `use_noqa` never gets set, resulting in it having no affect. `config.py` will need to be updated to set `use_noqa` in `get_validated_settings`.

PR #192, opened 2021-05-03
Reads use_noqa from config file.

lisroach:
## Summary
https://github.com/Instagram/Fixit/issues/191
`use_noqa` is not taking effect because the setting is not read from the config file, this diff fixes that by adding a `BOOLEAN_SETTINGS` option and looping through that.

`get_validated_settings` became too complicated (according to the linter) with this addition, so I separated it out into different functions.

## Test Plan
Added some unit tests, ran `tox`

### DIFF
--- a/fixit/common/tests/test_config.py
+++ b/fixit/common/tests/test_config.py
@@ -1,0 +1,57 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import itertools
+import os
+from pathlib import Path
+
+from libcst.testing.utils import UnitTest
+
+from fixit.common.config import (
+    BOOLEAN_SETTINGS,
+    LIST_SETTINGS,
+    NESTED_SETTINGS,
+    PATH_SETTINGS,
+    get_validated_settings,
+)
+
+
+TEST_CONFIG = {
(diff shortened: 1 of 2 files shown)


fixit/common/tests/test_config.py
```diff
@@ -0,0 +1,57 @@
...
+class TestConfig(UnitTest):
+    def test_get_validated_settings(self) -> None:
+        settings = get_validated_settings(TEST_CONFIG, Path("."))
+        settings_options = list(
+            itertools.chain(
+                BOOLEAN_SETTINGS, LIST_SETTINGS, PATH_SETTINGS, NESTED_SETTINGS
+            )
+        )
+        # Assert all the settings are set
+        self.assertTrue(all(setting in settings_options for setting in settings.keys()))
```
jimmylai, line 50: Can we also compare the setting values between settings_options and settings to verify they are the same?
```suggestion
        self.assertTrue(all(setting in settings_options for setting in settings.keys()))
```

jimmylai (approved):
Other than the comment, LGTM.

lisroach:
Updating the test made me realize the `file_contents` dict should be identical to the `settings` dict. I will adjust the JSON Schema PR (https://github.com/Instagram/Fixit/pull/188) to reflect this, since we won't need this code once the JSON Schema handles the type validation.

PR #198, opened 2021-06-04
Inheritable configs

lisroach:
## Summary
Putting up a draft to start discussion. This implements what was posted in #183. Configs are inheritable via simple hierarchy. There is nothing fancy like custom schema definitions or partial inheritance. A config will be inheritable if they add the key `inherit = True`, it defaults to `False` if the key is not set. 

Different keys are inherited differently, see `fixit/common/config.schema.json` for a breakdown and [jsonmerge](https://github.com/avian2/jsonmerge) for more details.

`allow_list_rules` is kind of a funny one, since if it is not included it defaults to all rules being allowed. If a leaf node inherits an empty `allow_list_rules` but the leaf config defines rules in the list, it has the effect of turning off all the rules that are not listed. 

`rule_config`s currently use jsonmerge's `objectMerge` with the defaults set. This means if there are two identical keys with different values, the value of the leaf config _overwrites_ the value of the root. If keys are not identical, they are both included in the final result. For example:

`root.fixit.config.yaml`
```
"rule_config": {
    "UnusedImportsRule": {
        "ignored_unused_modules": [
            "__future__",
            "__static__",
            "__static__.compiler_flags",
            "__strict__",
        ]
    },
   "Other Rule": { "yes": True}
}
```
`leaf.fixit.config.yaml`
```
"rule_config": {
   "UnusedImportsRule": {
        "ignored_unused_modules": ["__init__"]
  },
  "CoolRule": {"was": "sup"}
}
```
Results in:
```
"rule_config": {
   "UnusedImportsRule": {
        "ignored_unused_modules": ["__init__"]
  },
  "Other Rule": { "yes": True},
  "CoolRule": {"was": "sup"}
```

As expressed in the discussion post, I think this is most often the state that we want. But rules are variable, and I can forsee cases where someone would want this different (merging of lists instead of overwriting, for example, or completely overwriting). This seems like a good starting point for today.

## Test Plan
Added unit tests

### DIFF
--- a/fixit/common/base.py
+++ b/fixit/common/base.py
@@ -57,7 +57,7 @@ class LintConfig:
     packages: List[str] = field(default_factory=lambda: DEFAULT_PACKAGES)
     repo_root: str = "."
     rule_config: Dict[str, Dict[str, object]] = field(default_factory=dict)
-    inherit: bool = True
+    inherit: bool = False
 
 
 class BaseContext:
--- a/fixit/common/base.py
+++ b/fixit/common/base.py
@@ -57,6 +57,7 @@ class LintConfig:
     packages: List[str] = field(default_factory=lambda: DEFAULT_PACKAGES)
     repo_root: str = "."
     rule_config: Dict[str, Dict[str, object]] = field(default_factory=dict)
+    inherit: bool = True
 
 
 class BaseContext:
--- a/fixit/common/config.py
+++ b/fixit/common/config.py
@@ -17,9 +17,10 @@ import re
 from dataclasses import asdict
 from functools import lru_cache
 from pathlib import Path
-from typing import Any, Dict, Pattern, Set
+from typing import Any, Dict, List, Optional, Pattern, Set
 
 import yaml
+from jsonmerge import Merger
 from jsonschema import validate
 
 from fixit.common.base import LintConfig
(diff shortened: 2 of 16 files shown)


jreese:
More generally, is this expecting to up-front merge all configs, and then apply the same merged configs to all paths being linted? Or is this attempting to only find the hierarchy of configs relevant to each individual file, merge those configs, and repeat that find/merge individually for every single file being linted?

I would expect the latter to be the more "obvious/expected" implementation, but it looks like you're doing the former? If you're working on the former, and two subdirectories want to have two different sets of rules enabled or disabled, how is that conflict of desires being dealt with?

lisroach:
> is this expecting to up-front merge all configs, and then apply the same merged configs to all paths being linted?

No, that is not my intent.

> presence of a config in root/a/b.yaml would influence linting on root/d/f.py 

This also should not be the case, it walks up from the cwd to find all configs above, it should not walk back down a branch.

But these question do make me think about the current implementation. Right now, FixIt needs to be run from the bottom-most directory with a config file for that leaf config to be applied to all the files in its directory. I think if you run FixIt from a higher up directory it would miss the leaf configs. That isn't ideal, so I am taking a look now at the config file generation and how paths are found to be linted.

amyreese:

PR #202, opened 2021-08-05
Add dict comprehension to gather_sequential_await.py

yoonphillip:
## Summary

Calling an await within a dictionary comprehension runs it sequentially. Instead, this should typically be done using a gather call (i.e. gather_dict) to ensure that these will run in parallel. This checks that dict comprehensions also throw a Lint error.

## Test Plan

Added a unit test case and verified that all tests pass for this rule.

### DIFF
--- a/fixit/rules/gather_sequential_await.py
+++ b/fixit/rules/gather_sequential_await.py
@@ -59,6 +59,13 @@ class GatherSequentialAwaitRule(CstLintRule):
             """,
             line=2,
         ),
+        Invalid(
+            """
+            async def async_check_dict_comprehension():
+                {_i: await async_foo() for _i in range(0, 2)}
+            """,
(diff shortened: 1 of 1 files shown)


fixit/rules/gather_sequential_await.py
```diff
@@ -59,6 +59,13 @@ async def async_check_list_comprehension():
             """,
             line=2,
         ),
+        Invalid(
+            """
+            async def async_check_dict_comprehension():
+                {_i: await async_foo() for _i in range(0, 2)}
```
benhgreen, line 65: Should there also be test/handling for an async expression in the key?

yoonphillip:
Added test and support for await in dict key.

benhgreen (approved):

PR #204, opened 2021-11-19
Fix this usage: python3 -m fixit.cli.run_rules <path>

lpetre:
## Summary
Using a relative path here doesn't always work.

Consider this scenario:
 -  <path> is specified
 - it is relative to repo_root
 - but repo_root != cwd

In that case `fixit.cli.find_files` will locate zero files.

### DIFF
--- a/fixit/cli/args.py
+++ b/fixit/cli/args.py
@@ -112,7 +112,9 @@ def get_rules_parser() -> argparse.ArgumentParser:
 def relative_to_repo_root(_path: str) -> Path:
     repo_root = get_lint_config().repo_root
     try:
-        return Path(_path).resolve(strict=True).relative_to(repo_root)
+        path = Path(_path).resolve(strict=True)
+        path.relative_to(repo_root)
+        return path
     except ValueError:
         raise argparse.ArgumentTypeError(
             f"Invalid value {_path}.\n"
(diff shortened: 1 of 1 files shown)


fixit/cli/args.py
```diff
@@ -112,7 +112,9 @@ def get_rules_parser() -> argparse.ArgumentParser:
 def relative_to_repo_root(_path: str) -> Path:
     repo_root = get_lint_config().repo_root
     try:
-        return Path(_path).resolve(strict=True).relative_to(repo_root)
```
lpetre, line 115: You don't actually want to return a relative path here.

See the default here: https://github.com/Instagram/Fixit/pull/204/files#diff-1f90d45be47f4f2f218dff3e25fe7dc12f3cda5de358a8bfecafd92a9c89affbL130 

It is an absolute path.

These args are passed to fixit.cli.find_files, which is operating relative to CWD (not repo_root!)

https://github.com/Instagram/Fixit/blob/main/fixit/cli/__init__.py#L69-L72

lpetre: This method (relative_to_repo_root) was added in #84 

The intent is good: don't allow people to specify files outside of repo_root. That is why I still call relative_to below

zsol: Maybe we should change the name of the function. It doesn't return relative paths, it just validates that the path is under the repo root.

zsol (approved):

PR #207, opened 2021-11-20
Drop use of libcst.testing.*

lpetre:
## Summary
In order to upgrade pyre I'd like to stop using libcst.testing.utils.data_provider and use parameterized instead

The former attempts to typecheck the decorator input against the test method, but currently that is not possible in python's type system.

This is a large change w/ no functional difference. I used this codemod to convert all the tests: https://gist.github.com/lpetre/05b729a0891427c2a9def8d1eda3153d

I've changed from this:
```
from libcst.testing.utils import data_provider, UnitTest
class FooTest(UnitTest):
   @data_provider({"name": {"keyword": "value"}})
   def test(self, *, keyword: str) -> None:
      pass
```

To this equivalent in parameterized:
```
import unittest
from parameterized import param, parameterized
class FooTest(unittest.TestCase):
   @parameterized([param("name", **{"keyword": "value"})])
   def test(self, _name:str, keyword: str) -> None:
      pass
```

## Test Plan
python -m unittest

### DIFF
--- a/.pyre_configuration
+++ b/.pyre_configuration
@@ -6,6 +6,7 @@
         ".*/\\.tox/.*"
     ],
     "search_path": [
+        "stubs",
         {"site-package": "libcst"},
         {"site-package": "importlib_resources"},
         {"site-package": "flake8"}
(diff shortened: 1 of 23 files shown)


zsol:
Maybe add a type stub for parameterized, so pyre doesn't complain about a missing `expand` function?

amyreese:

PR #209, opened 2022-01-05
Upgrade pyre

lpetre:
This PR

 - upgrades pyre to 0.9.9 which brings a new type checking backend
- fixes some amount of type errors that 0.9.9 detects now
- suppresses the rest of the type errors

### DIFF
--- a/.pyre_configuration
+++ b/.pyre_configuration
@@ -5,5 +5,6 @@
     "exclude": [
         ".*/\\.tox/.*"
     ],
+    "workers": 3,
     "strict": true
 }
--- a/fixit/cli/__init__.py
+++ b/fixit/cli/__init__.py
@@ -216,9 +216,7 @@ def run_ipc(
     Returns an IPCResult object.
     """
 
-    resolved_paths = (
-        os.path.join(prefix, p) if prefix else p for p in paths
-    )
+    resolved_paths = (os.path.join(prefix, p) if prefix else p for p in paths)
 
     full_repo_metadata_config = opts.full_repo_metadata_config
     metadata_caches: Optional[Mapping[str, Mapping["ProviderT", object]]] = None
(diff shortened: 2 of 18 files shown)


.pyre_configuration
```diff
@@ -5,5 +5,6 @@
     "exclude": [
         ".*/\\.tox/.*"
     ],
+    "workers": 3,
```
lisroach, line 8: What's the Pyre bug this fixes?

lpetre: It is an issue where workers < 3 causes pyre to SIGSEGV. We discovered the issue (and the workaround) in this PR https://github.com/Instagram/LibCST/pull/579

lisroach: Yikes, thanks for the conext!

lisroach (approved):

fixit/cli/__init__.py
```diff
@@ -215,6 +217,7 @@ def run_ipc(
     Returns an IPCResult object.
     """
 
+    # pyre-fixme[35]: Target cannot be annotated.
     paths: Generator[str, None, None] = (
         os.path.join(prefix, p) if prefix else p for p in paths
     )
```
zsol, line 223: This should be easy to fix, it's just a name shadowing problem.

fixit/cli/apply_fix.py
```diff
@@ -144,6 +144,9 @@ def apply_fix_operation(
                         get_lint_config().formatter, updated_source
                     )
                 with open(path, "wb") as f:
+                    # pyre-fixme[6]: Expected `Union[array.array[typing.Any],
+                    #  bytearray, bytes, ctypes._CData, memoryview, mmap.mmap]` for 1st
+                    #  param but got `str`.
```
zsol, line 149: Oh cool, this (and the one below) are genuine bugs :) If `updated_source` has some non-ASCII characters, Fixit will clobber the file with invalid encoding.

We should open the file without binary mode:
```suggestion
                with open(path, "w") as f:
```

fixit/common/generate_pyre_fixtures.py
```diff
@@ -59,6 +59,7 @@ def gen_types_for_test_case(source_code: str, dest_path: Path) -> None:
         if "error" in data:
             raise PyreQueryError(cmd, data["error"])
         data = data["response"][0]
+        # pyre-fixme[35]: Target cannot be annotated.
```
zsol, line 62: This should be easy to fix too, just rename occurrences of `data` before this line to `raw_data`

PR #210, opened 2022-01-05
Update github actions to run on more python versions

lpetre:
Inspired by the CI actions in LibCST, this PR:
 - runs test, lint, and fixit in separate jobs
 - builds a distribution for pypi
 - runs on 3.9 and 3.10 python versions

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,125 +8,181 @@ on:
   pull_request:
 
 jobs:
+# Build the docs
   docs:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        python-version: [3.8]
-        os: [ubuntu-latest]
-
+    needs: lint
+    runs-on: ubuntu-latest
     steps:
-      - name: Checkout
-        uses: actions/checkout@v1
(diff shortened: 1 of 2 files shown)


lpetre:
I'll re-open this when my other PRs land

zsol (approved):

jreese:
Some suggestions:
- Use the new caching support in the `setup-python` action [[1][]]. That will save you a lot of work/configuration and make the whole file easier to read.
- Set up RTD to automatically build on PRs [[2][]], so you can skip the docs action entirely. I use this for µsort and all of my personal projects, and gives better signal because you can see the docs live on RTD.
- Why not run lint and typecheck on all Python runtimes? I've seen multiple cases in the past where lint fails on the oldest or newest runtime, but not others, for legitimate reasons that wouldn't have been caught otherwise. And at that point, maybe just combine those steps with the test step, so that it's fewer signals on the PR and easier to sort through.
- Same thing for coverage, as that will ensure that any `if sys.version_info < (3, 10)` blocks will still get counted towards final, combined coverage results.

[1]: https://github.blog/changelog/2021-11-23-github-actions-setup-python-now-supports-dependency-caching/
[2]: https://docs.readthedocs.io/en/stable/pull-requests.html

amyreese:

PR #212, opened 2022-01-06
CI isn't running in main because the branch was renamed

lpetre:
Running in main is important for coverage etc

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,7 @@ name: Build
 on:
   push:
     branches:
-      - master
+      - main
     tags:
       - v*
   pull_request:


.github/workflows/build.yml
```diff
@@ -2,7 +2,7 @@ name: Build
 on:
   push:
     branches:
-      - master
+      - main
     tags:
       - v*
```
lpetre, line 7: Is this really required? Can you push a tag that doesn't also push to the main branch?

lisroach: I don't know enough about github actions, but @jreese probably does

lisroach (approved):

.github/workflows/build.yml
jreese: Technically, yes, you can create a tag, and `git push --tags` or `git push origin <tagname>` to send the tag without updating main. There's also nothing in git that requires tags to point at commits reachable from named branches.

PR #214, opened 2022-01-07
Setuptools scm

lpetre:
## Summary
I've added setuptools_scm and configured it to generate fixit\_version.py

To improve our workflow around libcst releases, I'd like to propose that we publish pre-release versions of the package.

Draft until #210 lands

## Test Plan
CI


amyreese:

PR #216, opened 2022-02-11
Inheritable Configurations, Second Attempt

dkgi:
## Summary
A lot of this is copy pasting @lisroach's PR #198. Same motivation and 99% same implementation. 

We'd like configurations in subdirectories to be able to inherit from configurations in parent directories. We do this by setting `inherit: True` (defaults to `False`; probably discussion we should be having because I could see this lead to surprises). Got rid of some global state but currently have only implemented this for `run_rules`. Other commands might benefit from similar changes (but we'd have to have a discussion about what we can delete; there's a lot of code here that seems superfluous).

Implementation: the meat of this is making `get_lint_config` take a path as an argument. This defaults to `cwd` if none is given to be backwards compatible but realistically this shouldn't be an optional parameter. Also changed `get_config_rules` to `get_rules_for_path` and similarly added a path parameter. Rest of the change is adjusting to this (directly using in `run_rules` as well as fixing tests). 

Limitations: we're only using this to compute which rules to run at the moment. E.g. my understanding is that override options like `noqa` etc. Will be mostly ignored. Would prefer to do this as follow-ups to keep change size manageable.

## Test Plan
`tox -e py38`


fixit/cli/run_rules.py
```diff
@@ -1,8 +1,3 @@
-# Copyright (c) Facebook, Inc. and its affiliates.
-#
-# This source code is licensed under the MIT license found in the
-# LICENSE file in the root directory of this source tree.
-
```
zsol, line 5: I think we still need this

fixit/common/config.py
```diff
@@ -95,35 +90,69 @@ def get_validated_settings(
...
+    return schema
+
+
+CACHE: Dict[Path, LintConfig] = {}
+
+
+def get_lint_config(path: Optional[Path] = None) -> LintConfig:
+    """
+    Get configuration for linting a given path. If a path is provided we walk up
+    the path and merge configurations along the way until `inherit` is set to `False`.
```
zsol, line 116:
```suggestion
    Get configuration for linting a given path. If no path is given, default to CWD. Walk 
    up the path and merge configurations along the way until `inherit` is set to `False`.
```

fixit/common/config.py
```diff
@@ -95,35 +90,69 @@ def get_validated_settings(
...
 
-    cwd = Path.cwd()
-    for directory in (cwd, *cwd.parents):
-        # Check for config file.
-        # pyre-fixme[58]: `/` is not supported for operand types `object` and `Path`.
-        possible_config = directory / LINT_CONFIG_FILE_NAME
-        if possible_config.is_file():
-            with open(possible_config, "r") as f:
-                file_content = yaml.safe_load(f.read())
+            if isinstance(content, dict):
```
zsol, line 138: if there's a config file, but it's invalid (either because it's not yaml, or it's not a dict), we should assume `inherit` is the default value and `break`

This'd be easier to implement if we had a separate `get_default_settings` function which didn't take a `content` parameter

fixit/common/config.py
```diff
@@ -95,35 +90,69 @@ def get_validated_settings(
...
+    assert pkg, "No package was found, config types not validated."
+    config = pkg_resources.read_text(pkg, LINT_CONFIG_SCHEMA_NAME)
+    schema = json.loads(config)
+    return schema
+
+
+CACHE: Dict[Path, LintConfig] = {}
+
+
+def get_lint_config(path: Optional[Path] = None) -> LintConfig:
```
zsol, line 113: any reason for manually implementing that `CACHE` instead of applying `lru_cache`?

zsol (approved):
I agree that `inherit: True` as the default behavior would be less surprising even if it's a breaking change. I would change the default, document this change extensively, and make sure the next release bumps the major version.

But yeah, first, let's fix the limitations you mention in a separate PR

amyreese:

PR #217, opened 2022-05-13
Fix patched_source type hint and bug related to it

jack1142:
## Summary

Fixes a bug that caused apply_fix to write to a file in text mode rather than bytes mode even though it's trying to write bytes back.

## Test Plan

Run this in the root of the cloned repository of Fixit:
```
python -m venv .venv
.venv\Scripts\activate.bat
python -m pip install .
python -m fixit.cli.apply_fix
```
If using Linux/macOS, `. .venv/bin/activate` is needed for venv activation instead.


amyreese:

PR #221, opened 2022-07-10
Use jinja2<3.1 to fix docs build issue

jack1142:
## Summary

Fixes an incompatibility of the currently used Sphinx version with Jinja2 3.1. Simple fix until this project is ready to move to a newer version of Sphinx that supports Jinja2 3.1.

## Test Plan

Run `tox -e docs` in a clean environment (so without a `.tox` folder in the project's root directory)


amyreese:

PR #223, opened 2022-08-30
Config loading and hierarchical merging

amyreese:
## Summary

First piece of config loading based on the proposal in https://github.com/amyreese/Fixit/tree/config-prototype

- Walks the tree up from a target path to either a config with `root = true` or the top of the drive/filesystem.
- Loads all discovered files using tomllib/tomli.
- Basic priority-based overrides of values from outer configs.

Future work will add support for more intelligent merging of values rather than strict overrides, as well as support for multiple path-based overrides in a single config.

## Test Plan

Tests!

### DIFF
--- a/src/fixit/config.py
+++ b/src/fixit/config.py
@@ -7,6 +7,7 @@ import importlib
 import inspect
 import logging
 import pkgutil
+import sys
 from pathlib import Path
 from typing import Any, Collection, Dict, Iterable, List, Optional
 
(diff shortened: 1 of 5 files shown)


src/fixit/config.py
```diff
@@ -4,16 +4,111 @@
...
-def generate_config(path: Path) -> Config:
+
+def locate_configs(path: Path, root: Optional[Path] = None) -> List[Path]:
+    """
+    Given a file path, locate all relevant config files in priority order.
+
+    Walking upward from target path, creates a list of candidate paths that exist
+    on disk, ordered from nearest/highest priority to further/lowest priority.
+
+    If root is given, only return configs between path and root, ignoring any paths
```
zsol, line 24:
```suggestion
    If root is given, only return configs between path and root (inclusive), ignoring any paths
```

zsol (approved):

PR #225, opened 2022-08-30
use hatch for CI

zsol:


.github/workflows/build.yml
```diff
@@ -26,38 +26,16 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install
         run: |
-          pip install -U pip
-          pip install -Ur requirements-dev.txt
-          pip install .
+          pip install -U pip hatch
```
amyreese, line 29: Can we make hatch create the environment in this step, so that bootstrap/dependency problems show up here instead of in the test step?

ofek: Yup! https://hatch.pypa.io/latest/cli/reference/#hatch-env-create

PR #227, opened 2022-09-13
Rule testing

zsol:
# Enable testing for lint rules
The testing API is similar to fixit1, with a few changes:
1. `VALID` and `INVALID` are not CSTLintRule specific attributes
2. `VALID` and `INVALID` can now contain just strings, they'll be converted to `ValidTestCase` or `InvalidTestCase` automatically
3. BREAKING: `InvalidTestCase` doesn't accept a `line` and `column` attribute anymore. Instead there's a `range` attribute. This is to make assertions on violation ranges more accurate.
4. `expected_replacement` is not checked (yet).
5. test fixtures are not supported (yet).

This commit does a few things. It:
- adds a `scripts/migrate_rule_to_fixit2.py` script to make the necessary changes (transforming line/col to range in `InvalidTestCase`)
- adds a `fixit.testing` module to facilitate testing of rules
- uses this module in `tests/__init__.py` to load tests for rules
- updates some rules to make them pass


src/fixit/types.py
```diff
@@ -8,9 +8,11 @@
 from typing import Any, Dict, List, Optional
 
 from libcst._add_slots import add_slots
-from libcst.metadata import CodeRange
+from libcst.metadata import CodeRange as CSTCodeRange, CodePosition as CSTCodePosition
```
thatch: Why the alias? Is this to satisfy the typechecker to re-export? Would __all__ do the same thing?

src/fixit/__init__.py
```diff
@@ -11,10 +11,23 @@
...
-from .rule.cst import CSTLintRule, CstLintRule, InvalidTestCase, ValidTestCase
-from .types import Config, FileContent, Result
-
+from .rule.cst import CSTLintRule, CstLintRule
+from .testing import InvalidTestCase, ValidTestCase
+from .types import CodePosition, CodeRange, Config, FileContent, Result
 
 __all__ = [
     "__version__",
+    "fixit_bytes",
```
thatch, line 20: I understand this is for convenience, but when determining backwards compatibility is your intent to look at .api or this`__all__`?

thatch:
Is src/fixit/rules/use_fstring.py correct? I don't see code range there.

src/fixit/__init__.py
zsol: I suppose we could look at both, but maybe it makes sense to consider this `__all__` as the public api. Do you have opinions?

amyreese: I'm in favor of solidifying `__all__` as the public API, and having that be the list of things documented in an API reference. I would expect that any data types that we would think users will need when writing lint rules should be added to this list, so that there's a single location to import from.

thatch: Sounds fine. I've been waffling on this myself in other projects, mainly due to doing expensive imports up front if not needed for all parts of the api. I don't think that applies here.

amyreese (approved):
🤷‍♀️

src/fixit/types.py
zsol: I just wanted to make sure the name `CodeRange` is declared in this file explicitly instead of relying on an import. `__all__` would accomplish the same thing, yeah.

PR #229, opened 2022-10-06
Implementation of real configuration loading, merging, and overrides

amyreese:
## Summary

- [x] Reads `enable`, `disable`, and `options` values from config files
- [x] Reads `overrides` array of tables for configuring subdirectories
- [x] MVP "merging" of enable/disable rules based on sets with zero intelligence
- [x] Simple last-config-wins merging of options for individual rules
- [x] Basic validation of types, but not super thorough
- [x] Simple test cases for subdirectories, merging, and overrides
- [x] Documentation

## Test Plan

TESTS!!

DOCS!! https://fixit--229.org.readthedocs.build/en/229/guide.html

### DIFF
--- a/src/fixit/api.py
+++ b/src/fixit/api.py
@@ -4,9 +4,9 @@
 # LICENSE file in the root directory of this source tree.
 
 import logging
+import traceback
 from pathlib import Path
 from typing import Generator, Iterable, List
-import traceback
 
 import trailrunner
 
@@ -24,8 +24,7 @@ def _make_result(path: Path, violations: Iterable[LintViolation]) -> Iterable[Re
     except Exception as error:
         # TODO: this is not the right place to catch errors
         logger.debug("Exception while linting", exc_info=error)
-        tb = traceback.format_exc()
-        yield Result(path, violation=None, error=error, traceback=tb)
+        yield Result(path, violation=None, error=(error, traceback.format_exc()))
 
 
 def fixit_bytes(
@@ -57,8 +56,7 @@ def fixit_file(
 
     except Exception as error:
         logger.debug("Exception while fixit_file", exc_info=error)
-        tb = traceback.format_exc()
-        yield Result(path, violation=None, error=error, traceback=tb)
+        yield Result(path, violation=None, error=(error, traceback.format_exc()))
 
 
 def fixit_paths(
--- a/src/fixit/api.py
+++ b/src/fixit/api.py
@@ -6,6 +6,7 @@
 import logging
 from pathlib import Path
 from typing import Generator, Iterable, List
+import traceback
 
 import trailrunner
 
@@ -23,7 +24,8 @@ def _make_result(path: Path, violations: Iterable[LintViolation]) -> Iterable[Re
     except Exception as error:
         # TODO: this is not the right place to catch errors
         logger.debug("Exception while linting", exc_info=error)
-        yield Result(path, violation=None, error=error)
+        tb = traceback.format_exc()
+        yield Result(path, violation=None, error=error, traceback=tb)
 
 
 def fixit_bytes(
@@ -55,7 +57,8 @@ def fixit_file(
 
     except Exception as error:
         logger.debug("Exception while fixit_file", exc_info=error)
-        yield Result(path, violation=None, error=error)
+        tb = traceback.format_exc()
+        yield Result(path, violation=None, error=error, traceback=tb)
 
 
 def fixit_paths(
--- a/src/fixit/config.py
+++ b/src/fixit/config.py
@@ -230,7 +230,9 @@ def merge_configs(
         options: Optional[RuleConfigs] = None,
     ):
         subpath = subpath.resolve()
-        if not path.is_relative_to(subpath):
+        try:
+            path.relative_to(subpath)
+        except ValueError:  # not relative to subpath
             return
 
         for rule in enable:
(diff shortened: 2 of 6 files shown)


requirements-dev.txt
```diff
@@ -0,0 +1,8 @@
+black==22.8.0
+flake8==4.0.1
+flake8-bugbear==22.7.1
+mypy==0.971
+sphinx==4.3.2
+sphinx_mdinclude==0.5.1
+ufmt==2.0.1
+usort==1.0.5
```
zsol, line 8: Why have these? Shouldn't they live in `pyproject.toml`?

src/fixit/config.py
```diff
@@ -23,6 +40,12 @@
 log = logging.getLogger(__name__)
 
 
+class ConfigError(ValueError):
+    def __init__(self, msg: str, config: RawConfig):
```
zsol, line 44: Hmm, do we still get typechecking on this function without the return signature?
```suggestion
    def __init__(self, msg: str, config: RawConfig) -> None:
```

src/fixit/config.py
```diff
@@ -156,26 +224,82 @@ def merge_configs(
...
+    def process_subpath(
+        subpath: Path,
+        *,
+        enable: Sequence[str] = (),
+        disable: Sequence[str] = (),
+        options: RuleConfigs = frozendict(),
+    ):
+        subpath = subpath.resolve()
+        if not path.is_relative_to(subpath):
+            print(f"ignoring {subpath}")
```
zsol, line 242: this should either not be here, or converted into some kind of logging thing and suppressed by default

src/fixit/config.py
```diff
@@ -156,26 +224,82 @@ def merge_configs(
...
+            return
+
+        for rule in enable:
+            if rule.startswith("."):
```
zsol, line 246: IIUC, this hints at local rules being distinguished from packaged ones by a `.` prefix? I like this very much, wanna add it to [the proposal](https://github.com/amyreese/Fixit/tree/config-prototype)?

zsol: We should also maybe attach `local_path` to the `rule` to avoid collisions?
```
root/.fixit.toml  # has enable = [".foo"]
root/foo.py
root/subdir/.fixit.toml  # has enable = [".foo"]
root/subdir/foo.py
```

src/fixit/config.py
```diff
@@ -156,26 +224,82 @@ def merge_configs(
...
+        process_subpath(
+            config.path.parent,
+            enable=get_sequence(config, "enable"),
+            disable=get_sequence(config, "disable"),
+            options=get_options(config, "options"),
+        )
+
+        for override in get_sequence(config, "overrides"):
+            if not isinstance(override, dict):
+                raise ConfigError("'overrides' requires array of tables", config=config)
```
zsol, line 277: Hmm, maybe better to do validation in some other place? Like when we're loading the raw configs. It'd be also good to validate that `disable_rules` can't contain local rules (since they aren't supported by the code here)

src/fixit/types.py
```diff
@@ -62,3 +77,4 @@ class Result:
     path: Path
     violation: Optional[LintViolation]
     error: Optional[Exception] = None
+    traceback: Optional[str] = None
```
zsol, line 80: maybe merge these into `error: Optional[Tuple[Exception, Traceback]]`?

src/fixit/api.py
```diff
@@ -23,7 +24,8 @@ def _make_result(path: Path, violations: Iterable[LintViolation]) -> Iterable[Re
     except Exception as error:
         # TODO: this is not the right place to catch errors
         logger.debug("Exception while linting", exc_info=error)
-        yield Result(path, violation=None, error=error)
+        tb = traceback.format_exc()
+        yield Result(path, violation=None, error=error, traceback=tb)
```
zsol, line 28: maybe attach the traceback here and format at the place we're printing results?

zsol (approved):

requirements-dev.txt
amyreese: lol, oops, my commit testing thx on the project snuck in 😅

src/fixit/config.py
amyreese: I'm not sure about Pyre, but mypy has specific support for eliding the return type of `__init__` methods because they will never be anything else. See mypy#5677

src/fixit/api.py
amyreese: Does that play well with multiprocessing if the exc/tb is generated in the child and then formatted later in the parent?

src/fixit/config.py
amyreese: I agree that this is needed, but I'm going to leave this as-is for the moment, and think about how to do it right.

src/fixit/api.py
amyreese: I can't find a good way to type annotate traceback objects, and `type(tb)` gives me a type class that I can't use. Going to leave these as pre-formatted `str` for now, and we can revisit if you know the right way.

src/fixit/config.py
amyreese: Local rules in general aren't yet supported, but as a user, I would be disappointed if there weren't a way to disable local rules once they work.

src/fixit/api.py
zsol: I'm good with this for now, but have you tried https://docs.python.org/3/library/types.html#types.TracebackType ?

PR #230, opened 2022-10-06
Move dev dependencies into optional extras, fix RTD config

amyreese:
## Summary

- Moves dev dependencies into optional extras
- Configures hatch to install extras for relevant envs
- Configures RTD to use Py3.8 and install docs extras

## Test Plan

CI + RTD

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -2,8 +2,9 @@ version: 2
 sphinx:
   configuration: docs/conf.py
 python:
-  version: 3.7
+  version: 3.8
   install:
-    - requirements: requirements-dev.txt
     - method: pip
       path: .
+      extra_requirements:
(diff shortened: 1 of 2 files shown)


ofek:
It's very cool to see more users of Hatch environments 🙂

amyreese:
@ofek since you're here, is there a way (from the `pyproject.toml`) to change the path where hatch puts virtualenvs? I'd really like to have them in the project directory so that VSCode can more easily find them, and perhaps even symlink the default env to `.venv` so that VSCode finds it automatically.

ofek:
VSCode couldn't find them automatically until just now 🙂 https://hatch.pypa.io/latest/blog/2022/10/08/hatch-v160/#virtual-environment-location

PR #231, opened 2022-10-07
Update copyright headers, add lint script

amyreese:
## Summary

This uses the same copyright linter script that we use for µsort.

### DIFF
--- a/check_copyright.sh
+++ b/check_copyright.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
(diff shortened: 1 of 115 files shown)


check_copyright.sh
```diff
@@ -0,0 +1,16 @@
...
+
+set -eu
+
+die() { echo "$1"; exit 1; }
+
+while read filename; do \
+  grep -q "^# Copyright (c) Meta Platforms, Inc. and affiliates." "$filename" ||
+    die "Missing copyright in $filename"
+done < <( git ls-tree -r --name-only HEAD | grep "\(.py\|\.sh\)$" )
+
```
zsol, line 16: We should rewrite the copyright checker script in Python so people can run it on Windows

PR #240, opened 2022-11-02
FP2: Local Lint Rules

### LINKED ISSUE #237: FP2: Local Rules
Proposal: soon

amyreese:
## Summary

Draft proposal for managing local lint rules: https://fixit--240.org.readthedocs.build/en/240/proposals/fp2.html

Closes #237

### DIFF
--- a/docs/proposals/fp2.rst
+++ b/docs/proposals/fp2.rst
@@ -5,7 +5,8 @@ By default, Fixit expects rules to be referenced by their fully-qualified
 module and/or class name. For example, ``fixit.rules`` refers to the bundled set
 if lint rules shipped with Fixit. Third party rules available in the environment
 can similarly be referenced by their module name, as long as they are installed
-and importable by Fixit at runtime.
+and importable by Fixit at runtime. These will be collectively referred to as
+"global rules" here for sake of clarity.
 
 However, in many cases, it is also useful for projects using Fixit to build
 custom lint rules specific to the project being linted. It is beneficial for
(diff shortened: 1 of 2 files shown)


docs/proposals/fp2.rst
```diff
@@ -0,0 +1,145 @@
...
+to global rules:
+
+.. code-block:: toml
+
+    [tool.fixit]
+    enable = [".rules"]
+
+    [[tool.fixit.overrides]]
+    path = "project1"
+    disable = [".rules.PickyRule"]
```
thatch, line 48: AIUI inheritable configs mean this as well as the parent might be able to `enable = [".rules"]` -- if that's correct, which one would this disable?

docs/proposals/fp2.rst
```diff
@@ -0,0 +1,145 @@
...
+
+---
+
+When discovering and loading rules, the system should attempt to make sure
+that it is loading rules from the local path, rather than accidentally loading
+rules from the outside environment. This can be done either by a temporarily
+restricted path when importing, or with a custom import loader.
+
+The former may be a simpler starting point, and can be handled with a temporary
+override of ``sys.path`` to include the local directory — and nothing else? —
```
thatch, line 80: For a prototype, this is fine, but we'll need to find a better way.

1. you can't import any name that has already been imported (e.g. `enable = [".sys"]` will not work
2. if two levels have the same name (like ".lints" let's say) they will collide and be reused
3. You won't be able to import outside of that dir (this might be on purpose, but will encourage them to do sys.path hackery in the linters, and isn't explicitly documented here)

I don't see anything in importlib with custom loaders that does exactly what we'd want, so I will look in e.g. pytest or something else that imports plugins to see if there is some API I'm missing.

docs/proposals/fp2.rst
```diff
@@ -0,0 +1,145 @@
...
+    path = "project1"
+    disable = [".rules.PickyRule"]
+    enable = [".project1.rules"]
+
+
+Implementation
+--------------
+
+When gathering enabled/disabled rules from configuration, Fixit currently
+just records the module name (fqdn) and does set operations when merging
```
thatch, line 56: s/fqdn/qualname/

docs/proposals/fp2.rst
```diff
@@ -0,0 +1,145 @@
...
+The former may be a simpler starting point, and can be handled with a temporary
+override of ``sys.path`` to include the local directory — and nothing else? —
+and then importing the module as normal, with the leading period removed.
+
+This could look something like:
+
+.. code-block:: python
+
+    with temporary_sys_path(parent_path):
+        name = name.lstrip(".")
```
thatch, line 88: What should two leading dots do?

docs/proposals/fp2.rst
thatch, line 48: Additionally, this is from FP1, but why not `.rules:PickyRule` with a colon to be more like PEP 517? If that's not the original location where PickyRule is defined, can I assume it does a lookup of `PickyRule.__module__` to determine a canonical path?  Is it an error to disable something that fixit doesn't think is enabled?

amyreese: `".rules"` would refer to the same thing within a single file (ie, local rules in overrides are not affected by the subpath). So the top-level `enable = [".rules"]` and the override's `enable = [".rules"]` are enabling the exact same local rules module, and therefore the `disable = [".rules.PickyRule"]` is similarly deterministic.

docs/proposals/fp2.rst
amyreese: The "limitations" section mentions that we want to specifically disallow things like that. I expect that we'll have some validation when reading configs that calls that out as not allowed.

docs/proposals/fp2.rst
amyreese: I think I would personally prefer to use `foo.bar:Baz`, but `foo.bar.Baz` is what Fixit has used for a while, and basically just tries to import `foo`, then walk the dotted parts to find the final result.

If we switch to `foo.bar:Baz`, then we could just assume that `foo.bar` will always be the full module name, and import that directly rather than needing to walk modules/objects and assume/hope that everything is imported appropriately.

@zsol, since we have the opportunity to change this, do you have an opinion?

>  Is it an error to disable something that fixit doesn't think is enabled?

No, it just goes in the set of things to disable, and if that thing never got enabled, then nbd.

thatch: I don't know that we're talking about the same thing.  Let's say we have

```
fixit.toml
rules/GoodRule.py
rules/PickyRule.py
subdir/fixit.toml
subdir/rules/ThirdRule.py
```

Is there a way for toplevel fixit.toml to enable `.rules` and subdir to disable toplevel's `.rules.PickyRule`?  I am interpreting this text from FP1 when assuming that toplevel's enables would be inherited, with emphasis added: "or by _separate config files within those subpaths_. Overrides should be applied to all files within the relative subpaths, accounting for further nested overrides."

docs/proposals/fp2.rst
thatch: And 4. rules won't be able to use relative imports within themselves.

Item 4 specifically can be fixed with [importlib.import_module](https://docs.python.org/3/library/importlib.html#importlib.import_module) specifying `package` as the one containing the toml file.  I don't _think_ this will execute intermediate `__init__.py` (which might be problematic in a megarepo), and helps with 1 and 2 but maybe makes 3 make more sense but encourage less explicit dependencies.

docs/proposals/fp2.rst
zsol: I'm happy to do `foo.bar:Baz`, it's both more clear to the reader and simpler to implement.

amyreese: No, the proposal is that a config in a subdirectory would not be able to disable local rules defined in a parent directory's config; the suggested method is either to use overrides in that parent config, or the escape hatch of setting `root=true` and abandoning the parent configs entirely.

This is mainly because I couldn't imagine a reasonable way to define exactly what level of local rules are being enabled/disabled without getting into weird syntax or structure that ended up being more complicated to document and implement. If you have suggestions for how to do this without ambiguity, please share. :)

PR #245, opened 2022-12-06
Enable fixit as linter for project

amyreese:
* #252
* #251
* #250
* #249
* __->__ #245
* #244

- Enables the default set of rules
- Disables the frozen dataclass and noqa rules (for now?)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,6 +75,7 @@ typecheck = "mypy --install-types --non-interactive src/fixit"
 [tool.hatch.envs.lint.scripts]
 check = [
     "flake8 src/fixit",
+    "fixit lint src/fixit",
     "ufmt check src/fixit",
     "{root}/check_copyright.sh",
 ]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -92,4 +93,7 @@ build = [
 
 [tool.fixit]
 enable = ["fixit.rules"]
-disable = ["fixit.test.foo", "fixit.test.bar"]
+disable = [
+    "fixit.rules:ExplicitFrozenDataclassRule",
+    "fixit.rules:UseLintFixmeCommentRule",
```
zsol, line 98: maybe add a comment explaining the reason for disabling these? Feels weird to ship a lint rule as part of fixit but not enable it on fixit itself :)

zsol (approved):

pyproject.toml
amyreese: Sure thing. These feel like *very* opinionated rules, and I would like to have a discussion about moving them (and maybe others?) into an "extras" subpackage that aren't selected by default (with the new collection behavior), or maybe extracting them into a separate package/plugin altogether.

PR #250, opened 2022-12-08
Implement local rule loading

amyreese:
* #252
* #251
* __->__ #250
* #249
* #245
* #244

- Refactors module collection and walking into top level functions
- Adds new helper context manager to allow importing arbitrary local
  rules as if they were in the `fixit.local` namespace
- Imports local rules directly from qualified namespace

### DIFF
--- a/src/fixit/config.py
+++ b/src/fixit/config.py
@@ -8,6 +8,7 @@ import inspect
 import logging
 import pkgutil
 import sys
+from contextlib import contextmanager
 from dataclasses import replace
 from pathlib import Path
 from types import ModuleType
@@ -16,6 +17,7 @@ from typing import (
     Collection,
     Dict,
     Iterable,
+    Iterator,
     List,
     Mapping,
     Optional,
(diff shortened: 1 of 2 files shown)


amyreese:
I don't have good test cases, but I specifically validated that this approach allows importing distinct `.local.copyright` modules from multiple hierarchies with nested `fixit.toml` configs. Will try to build a good test case soon.

src/fixit/config.py
```diff
@@ -58,71 +60,147 @@ def __init__(self, msg: str, rule: QualifiedRule):
...
+        import fixit.local
+
+        assert hasattr(fixit.local, "__path__")
+        assert fixit.local.__spec__ is not None
+        assert rule.root is not None
+
+        orig_spec = fixit.local.__spec__
+        fixit.local.__path__ = [rule.root.as_posix()]
+        fixit.local.__spec__ = importlib.machinery.ModuleSpec(
+            name="fixit.local",
```
zsol, line 102: maybe worth factoring `"fixit.local"` into a global variable

src/fixit/local/__init__.py
```diff
@@ -0,0 +1,6 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Fake package to serve as module root when loading local rules
```
zsol, line 6: nit
```suggestion
"""Fake package to serve as module root when loading local rules"""
```

zsol (approved):
Neat :)

amyreese:
Manually merged and pushed to main because ghstack isn't happy.

PR #218, opened 2022-05-13
Add pre-commit hooks configuration

jack1142:
## Summary

Fixes #193 
Since I'm not very familiar with how this tool works, I am unsure whether it is safe to run multiple Fixit processes (which is what pre-commit does by default by splitting out passed file names evenly) in parallel or if Fixit should be passed with all files at once. If it can cause some issues, I can add [`require_serial` option](https://pre-commit.com/#hooks-require_serial).

It would be great also to include this in the documentation though I don't know where you would want me to put it, it doesn't seem like there's any document related to integrations with IDEs and other tooling yet.

## Test Plan

1. Install pre-commit using:
```
python -m pip install pre-commit
```
2. Create a repository with a lint rule and autofixer to test on (I'm not familiar with this tool so I can't really give a specific example, I assume a reviewer will be able to come up with one)
3. Run pre-commit on all files to test the behavior of the hook:
```
pre-commit run --all-files
```

PR #254, opened 2022-12-15
Add pre-commit hooks configuration

### LINKED ISSUE #193: Pre-commit hook configuration
It would be awesome to get out-of-the-box support for [pre-commit](https://pre-commit.com/) as plenty of other linters and auto-formatters such as flake8, pylint, black, or isort provide.

This is a matter of creating an appropriate `.pre-commit-hooks.yaml` file in the root of the repository as shown here: https://pre-commit.com/#new-hooks

Jackenmen:
## Summary

Supersedes #218, fixes #193
Since I'm not very familiar with how this tool works, I am unsure whether it is safe to run multiple Fixit processes (which is what pre-commit does by default by splitting out passed file names evenly) in parallel or if Fixit should be passed with all files at once. If it can cause some issues, I can add [`require_serial` option](https://pre-commit.com/#hooks-require_serial).

I added a new document for integrations as it seems like a good place to also document editor integrations (or perhaps CI integrations) in the future and I felt that the user guide would probably become too crowded in the future to use for documentation anyway.

## Test Plan

1. Install pre-commit using:
```
python -m pip install pre-commit
```
2. Create a repository with a lint rule and autofixer to test on (I'm not familiar with this tool so I can't really give a specific example, I assume a reviewer will be able to come up with one)
3. Run pre-commit on all files to test the behavior of the hook:
```
pre-commit run --all-files
```

### DIFF
--- a/docs/integrations.rst
+++ b/docs/integrations.rst
@@ -1,8 +1,8 @@
 Integrations
 ============
 
-Version control integration (pre-commit)
-----------------------------------------
+pre-commit
+----------
 
 Fixit can be included as a hook for `pre-commit <https://pre-commit.com>`__.
 
--- a/docs/integrations.rst
+++ b/docs/integrations.rst
@@ -1,0 +1,31 @@
+Integrations
+============
+
+Version control integration (pre-commit)
+----------------------------------------
+
+Fixit can be included as a hook for `pre-commit <https://pre-commit.com>`__.
+
+Once you `install it <https://pre-commit.com/#installation>`__, you can add Fixit's pre-commit hook to the ``.pre-commit-config.yaml`` file in your repository.
+
+- To run lint rules on commit, add:
+
+.. code:: yaml
+
+    repos:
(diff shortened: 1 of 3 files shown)


amyreese:
It shouldn't be a "problem" to run multiple instances in parallel, though it would potentially be suboptimal, as the new version of Fixit has builtin support for multiprocessing, and running multiple versions in parallel a) would require eating the cost of Python startup for each file, and b) potentially result in "too many" child processes if/when we further parallelize multiple lint rules on a single file. 

That said, I think it's fine with defaults for now, and we can revisit that choice later down the road if it gets out of hand.

docs/integrations.rst
```diff
@@ -0,0 +1,31 @@
+Integrations
+============
+
+Version control integration (pre-commit)
```
amyreese, line 4: Maybe just "pre-commit" for the section title?

amyreese (approved):

amyreese:
Thank you for contributing this :)

PR #257, opened 2023-02-08
Formatting the linting output

Chenguang-Zhu:
## Summary
Currently, Fixit just dumps objects to stdout. This pull request changes the output  to this format: `<filepath>@<line>:<col> <rulename>: <summary>`. It also colors the output in red. 

## Test Plan
`make test lint html` should give no errors. 
Run `hatch env run -- fixit lint .` in the Fixit root directory, the new output format is as follows:
Screenshot 2023-02-07 at 4 05 12 PM

### DIFF
--- a/src/fixit/api.py
+++ b/src/fixit/api.py
@@ -19,7 +19,10 @@ from .ftypes import Config, FileContent, LintViolation, Result
 logger = logging.getLogger(__name__)
 
 
-def print_result(result: Result, debug: bool) -> None:
+def print_result(result: Result, debug: bool = False) -> None:
+    """
+    Print linting results in a clear format for easy understanding
+    """
     path = result.path
     if result.violation:
         rule_name = result.violation.rule_name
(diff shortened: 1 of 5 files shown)


amyreese (changes requested):
Just needs to handle the case where `error` is set, eg, when an exception occurred while processing a file. The `error` attribute contains both the exception object and a string containing the traceback information. In the normal case, I think we would skip the traceback, and maybe only print it if the `--debug` flag has been passed. Also, I think it could make sense to have normal lint violations print in yellow, but errors print in red.

I think it might be better to make the `__str__` method a separate `print_result` function that takes a `Result` object and returns a string. That would not only allow for passing an alternative (eg `json_result`) but would also allow passing options like the `--debug` flag mentioned above.

Chenguang-Zhu:
Hi @amyreese , I updated the code to handle error cases. Now the violations are in the following format:
Screenshot 2023-02-08 at 5 53 59 PM

When the file has errors, without `--debug`:
Screenshot 2023-02-08 at 5 55 11 PM

With `--debug`:
Screenshot 2023-02-08 at 5 58 23 PM

src/fixit/api.py
```diff
@@ -17,6 +19,27 @@
...
+        rule_name = result.violation.rule_name
+        start_line = result.violation.range.start.line
+        start_col = result.violation.range.start.column
+        message = result.violation.message
+        click.secho(
+            f"{path}@{start_line}:{start_col} {rule_name}: {message}", fg="yellow"
+        )
+    else:
+        # An exception occurred while processing a file
+        if result.error:
```
amyreese, line 34: Better to combine this into a single `elif` rather than nesting.

src/fixit/api.py
```diff
@@ -17,6 +19,27 @@
...
+            error = result.error[0]
+            traceback_info = result.error[1]
```
amyreese, line 36: Prefer simple unpacking, like:

    error, tb = result.error

src/fixit/api.py
```diff
@@ -17,6 +19,27 @@
 logger = logging.getLogger(__name__)
 
 
+def print_result(result: Result, debug: bool) -> None:
```
amyreese, line 22: Set a default value for `debug = False`, to make this a nicer API for external usage.

Would also appreciate a docstring here, and add it to the "simple API" sections of the sphinx docs (see `docs/api.rst`).

amyreese (approved):
Overall looks good. Just have a few outstanding nits left before we can merge this. Cheers!

PR #259, opened 2023-02-18
Enable FullRepoManager when lint rules need it

amyreese:
* (to be filled)

Looks at the list of enable lint rules, and if any of those need the
`FullyQualifiedNameProvider` metadata, and then enables the
`FullRepoManager` from libcst. The cache is then resolved and passed
through the metadata wrapper when visiting the CST, enabling rules to
use the appropriate metadata.

Requires passing the root path and file path being linted to
`FullRepoManager`, which is accomplished by passing the Fixit config
further through the API into the `CSTLintRunner` implementation.

### DIFF
--- a/src/fixit/rule/cst.py
+++ b/src/fixit/rule/cst.py
@@ -17,6 +17,7 @@ from typing import (
     Iterator,
     Mapping,
     Optional,
+    Set,
     Union,
 )
 
@@ -30,12 +31,14 @@ from libcst import (
 from libcst.metadata import (
     CodePosition,
     CodeRange,
+    FullRepoManager,
+    FullyQualifiedNameProvider,
     MetadataWrapper,
     PositionProvider,
     ProviderT,
 )
 
-from fixit.ftypes import FileContent, LintViolation
+from fixit.ftypes import Config, FileContent, LintViolation
 from . import LintRule, LintRunner, TimingsHook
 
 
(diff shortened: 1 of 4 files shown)


amyreese:
Depends on changes in https://github.com/Instagram/LibCST/pull/867 to correctly handle passing absolute paths to `FullRepoManager`.

src/fixit/rule/cst.py
```diff
@@ -69,10 +73,26 @@ def visit_hook(name: str) -> Iterator[None]:
                 logger.debug(f"PERF: {name} took {duration_us} µs")
                 self.timings[name] += duration_us
 
+        metadata_cache: Mapping[ProviderT, object] = {}
+        needs_repo_manager: Set[ProviderT] = set()
+
         for rule in rules:
             rule._visit_hook = visit_hook
+            if FullyQualifiedNameProvider in rule.METADATA_DEPENDENCIES:
```
zsol, line 81: This is not enough, we need to inspect transitive dependencies of the rule. You'll want to call `get_inherited_depedencies()`: https://py.wtf/libcst/libcst._metadata_dependent/MetadataDependent#libcst._metadata_dependent.MetadataDependent.get_inherited_dependencies

zsol: sidenote: it sucks that there's no good way to inspect metadataproviders to see if they need a `FullRepoManager` (today the best way is to check if they have a `gen_cache()` classmethod)

zsol (changes requested):

amyreese:
Updated to gather transitive dependencies and check for gen_cache, so that TypeInferenceProvider will also be supported.

zsol (approved):

PR #271, opened 2023-03-08
Test rules in `pyflakes` directory

ansley:
* #277
* #276
* #275
* #274
* #273
* #272
* __->__ #271

### DIFF
--- a/docs/guide.rst
+++ b/docs/guide.rst
@@ -6,10 +6,10 @@ Configuration
 
 Fixit uses `TOML format <https://toml.io>`_ for configuration, and supports
 hierarchical, cascading configuration. Fixit will read values from both the
-standardized ``pyproject.toml`` file as well as a separate ``fixit.toml`` file,
-with values from the latter taking precendence over the former, and values from
-files "nearer" to those being linted taking precedence over values from files
-"further" away.
+standardized ``pyproject.toml`` file as well as a separate ``.fixit.toml`` or
+``fixit.toml`` file, with values from the latter taking precendence over the
+former, and values from files "nearer" to those being linted taking precedence
+over values from files "further" away.
 
 When determining the configuration to use for a given path, Fixit will continue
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -72,7 +72,7 @@ version-file = "src/fixit/__version__.py"
 features = ["dev", "pretty"]
 
 [tool.hatch.envs.default.scripts]
-test = "python -m fixit.tests"
+test = "python3 -m fixit.tests"
```
amyreese, line 75: Why was this change needed?  Hatch environments should be controlled enough that just `python` will always work.

src/fixit/tests/__init__.py
```diff
@@ -16,3 +16,7 @@
     globals(),
     collect_rules(enables=[QualifiedRule("fixit.rules")], disables=[]),
 )
+add_lint_rule_tests_to_module(
+    globals(),
+    collect_rules(enables=[QualifiedRule("fixit.rules.pyflakes")], disables=[]),
```
amyreese, line 21: This breaks the tests since the module doesn't (yet) exist. Let's go ahead and create an empty `__init__.py` there so that the tests can pass here.

amyreese (changes requested):

src/fixit/tests/__init__.py
amyreese: You can leave these lines in, and then add `src/fixit/rules/__init__.py`. That will provide the "fixit.rules.pyflakes" package that these lines are telling it to look for.

The following diffs in the stack accomplish the same thing, but it would be nice to have CI passing on all diffs in the stack, and the empty init will fill that role.

pyproject.toml
amyreese: I think this being "python3" is what's causing the CI failures on windows. My guess is that the virtualenv doesn't have a python3.exe in it, and so it's finding a random non-venv runtime, and therefore can't find/import fixit.

PR #272, opened 2023-03-08
[F402] Warn on loop variable shadowing module name

ansley:
* #277
* #276
* #275
* #274
* #273
* __->__ #272
* #271


amyreese (changes requested):
Needs formatting fixes (`make format` or `hatch run lint:fix`).

amyreese:
I think we need more variety of test cases, covering things like loops with multiple elements (or deep names), as well as some validating things like `import a.b` where `a` is the name that might be shadowed, but `b` isn't.  Take a look at upstream pyflakes tests (search for "shadow" to find instances), and see what other cases we can cover from their examples.

amyreese (approved):

PR #273, opened 2023-03-08
[F403] Warn on star import

ansley:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #277
* #276
* #275
* #274
* __->__ #273
* #272
* #271

PR #274, opened 2023-03-08
[F403, F405, F821] Warn on undefined names

ansley:
* #277
* #276
* #275
* __->__ #274
* #273
* #272
* #271


src/fixit/rules/pyflakes/no_undefined_name.py
```diff
@@ -0,0 +1,143 @@
...
+                print(i)
+            """
+        ),
+        Valid(
+            """
+            def fn():
+                global b
+                b = 3
+
+            x = b + 2
```
amyreese, line 104: where is `b` defined here?

PR #275, opened 2023-03-08
[F811] Warn on and delete double import

ansley:
* #277
* #276
* __->__ #275
* #274
* #273
* #272
* #271

### DIFF
--- a/docs/guide.rst
+++ b/docs/guide.rst
@@ -6,10 +6,10 @@ Configuration
 
 Fixit uses `TOML format <https://toml.io>`_ for configuration, and supports
 hierarchical, cascading configuration. Fixit will read values from both the
-standardized ``pyproject.toml`` file as well as a separate ``fixit.toml`` file,
-with values from the latter taking precendence over the former, and values from
-files "nearer" to those being linted taking precedence over values from files
-"further" away.
+standardized ``pyproject.toml`` file as well as a separate ``.fixit.toml`` or
+``fixit.toml`` file, with values from the latter taking precendence over the
+former, and values from files "nearer" to those being linted taking precedence
+over values from files "further" away.
 
 When determining the configuration to use for a given path, Fixit will continue
 looking upward in the filesystem until it reaches either the root of the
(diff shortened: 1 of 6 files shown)


src/fixit/rules/pyflakes/no_double_import.py
```diff
@@ -0,0 +1,210 @@
...
+        # handle the situation. We should add additional logic for this special
+        # case or create a completely different lint rule.
+        Invalid(
+            """
+            import a
+            from x import a
+            """,
+            expected_replacement="""
+            import a
+            """,
```
amyreese, line 199: For these shadowed imports, this autofix is incorrect — it will result in different runtime behavior than what the original code was, because the last import of a name overwrites all previous imports of that name.

If we can't get that right, then we need to skip the autofix, or defer this rule until a future time when we can do it correctly.

src/fixit/rules/pyflakes/no_double_import.py
```diff
@@ -0,0 +1,210 @@
...
+            x = g.b([1, 3, 5, 6, 8])
+            """,
+        ),
+        Invalid(
+            """
+            import a, b, a
+            """,
+            expected_replacement="""
+            import a, b
+            """,
```
amyreese, line 177: I know in this case, both `a` imports are referring to the same module, but I don't like the idea that we're removing the *second* import. We should instead be removing the first one, so that we're giving consistent runtime behavior before/after. For imports with side-effects, order unfortunately matters.

src/fixit/rules/pyflakes/no_double_import.py
```diff
@@ -0,0 +1,210 @@
...
+        ),
+        Invalid(
+            """
+            import a, b
+            import a
+            """,
+            expected_replacement="""
+            import a, b
+            """,
```
amyreese, line 186: same

src/fixit/rules/pyflakes/no_double_import.py
```diff
@@ -0,0 +1,210 @@
...
+        ),
+        Invalid(
+            """
+            from x import a
+            from y import a
+            """,
+            expected_replacement="""
+            from x import a
+            """,
```
amyreese, line 208: same

amyreese (changes requested):

PR #276, opened 2023-03-08
[F812] Warn when list comprehension redefines a name

ansley:
* #277
* __->__ #276
* #275
* #274
* #273
* #272
* #271

This lint rule is probably in Flake8 to prevent elements from leaking
out of the comprehension scope, but the behavior was fixed in Python3.
I worry such a strict lint rule will cause more problems than it solves,
considering we use Python 3.10 at Meta. Perhaps we should consider
disabling this rule by default.

### DIFF
--- a/docs/guide.rst
+++ b/docs/guide.rst
@@ -6,10 +6,10 @@ Configuration
 
 Fixit uses `TOML format <https://toml.io>`_ for configuration, and supports
 hierarchical, cascading configuration. Fixit will read values from both the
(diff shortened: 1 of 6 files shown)


amyreese (changes requested):
Can you add some test cases covering mutually exclusive scopes, where I would expect this to be allowed?

Something like:

```py
def foo():
    i = 1
    print(i)

def bar():
    values = [i * 2 for i in range(5)]
```

PR #279, opened 2023-03-10
Basic multiprocessing when linting multiple files

amyreese:
* #281
* __->__ #279

Enables simple concurrency when linting multiple files using the
multiprocessing implementation from trailrunner. Each file will be
linted in its own child process, reducing runtime against the fixit
codebase by approximately 60%.

Future performance improvements to consider would focus around
classifying fast/slow rules, and running slower rules as separate jobs.

### DIFF
--- a/src/fixit/api.py
+++ b/src/fixit/api.py
@@ -83,6 +83,14 @@ def fixit_file(
         yield Result(path, violation=None, error=(error, traceback.format_exc()))
 
 
+def _fixit_file_wrapper(path: Path) -> List[Result]:
+    """
+    Wrapper because generators can't be pickled or used directly via multiprocessing
+    TODO: replace this with some sort of queue or whatever
+    """
+    return list(fixit_file(path))
+
+
 def fixit_paths(
     paths: Iterable[Path],
 ) -> Generator[Result, None, None]:
@@ -96,5 +104,8 @@ def fixit_paths(
     for path in paths:
         expanded_paths.extend(trailrunner.walk(path))
 
-    for path in expanded_paths:
-        yield from fixit_file(path)
+    if len(expanded_paths) == 1:
+        yield from fixit_file(expanded_paths[0])
+    else:
+        for path, results in trailrunner.run_iter(expanded_paths, _fixit_file_wrapper):
+            yield from results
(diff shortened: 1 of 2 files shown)


src/fixit/api.py
```diff
@@ -96,5 +104,8 @@ def fixit_paths(
     for path in paths:
         expanded_paths.extend(trailrunner.walk(path))
 
-    for path in expanded_paths:
-        yield from fixit_file(path)
+    if len(expanded_paths) == 1:
+        yield from fixit_file(expanded_paths[0])
+    else:
+        for path, results in trailrunner.run_iter(expanded_paths, _fixit_file_wrapper):
```
zsol, line 110:
```suggestion
        for _, results in trailrunner.run_iter(expanded_paths, _fixit_file_wrapper):
```

zsol (approved):

src/fixit/api.py
```diff
@@ -96,5 +104,8 @@ def fixit_paths(
     for path in paths:
         expanded_paths.extend(trailrunner.walk(path))
 
-    for path in expanded_paths:
-        yield from fixit_file(path)
+    if len(expanded_paths) == 1:
```
thatch, line 107: Up to you whether to specialcase here, but I like that trailrunner _doesn't_.

I think we've had some offline discussions about the value of "universality" of bugs when coming up with [usort's goals](https://usort.readthedocs.io/en/stable/why.html), that if you're adding test cases for coverage's sake maybe just remove the branches.

> A single primary code path, to reduce the amount of code and number of test cases needed to guarantee safety.

This (spawn in multiprocessing issues, specifically) has come up several times in the past, and this might paper over ones like https://github.com/facebookincubator/Bowler/issues/52 (only on platforms we didn't use ourselves) or the reason that [dropping 3.6 support](https://github.com/omnilib/trailrunner/commit/466b71e42a71b0c38ac541f7e6f6b64d4ef136ab) didn't get more objection from me (because the primary alternative was only use a ThreadPoolExecutor there, which is icky).

Is the fast/slow classification relatively soon, and this is just a temporary state?

amyreese: The main reason/concern for special casing is to reduce the startup cost when linting a single file. We have the same special cases in µsort and µfmt for the same reason. I agree that it defeats the "purity", but having appropriate test cases (that don't just use a thread pool) and/or functional tests helps to make sure that we're exercising the multiprocessing codepaths as well.

> Is the fast/slow classification relatively soon, and this is just a temporary state?

Probably not "soon", and I don't expect this special case to be temporary, due to the given performance reasons above.

PR #284, opened 2023-03-17
Remove unused Sphinx inventories

michel-slm:
## Summary
Remove unused inventories from `docs/conf.py`.

These are not used, and `stdlib_list` has not been updated since 2015.

## Test Plan
After removing, documentation builds fine for the upcoming update of the Fedora package.


amyreese:
Cheers

PR #297, opened 2023-04-07
new rule - use async sleep in async def

jiayiyaij:
asyncio.sleep whould be used in async functions. This rule is to capture if time.sleep is used in an async function

### DIFF
--- a/docs/guide/builtins.rst
+++ b/docs/guide/builtins.rst
@@ -32,6 +32,7 @@ unless explicitly listed in the :attr:`disable` configuration option.
 - :class:`SortedAttributesRule`
 - :class:`UseAssertInRule`
 - :class:`UseAssertIsNotNoneRule`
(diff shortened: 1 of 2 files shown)


amyreese:
@jiayiyaij did you consider using `libcst.metadata.FullyQualifiedNameProvider` for this? I think it could reduce or remove the need for visiting and tracking imports manually, by simply asking the metadata provider for the qualified name of each call, and then seeing if that is either `time.sleep` or `asyncio.sleep`?  It might not catch the star imports, but maybe that's not as big a deal since we plan to throw lint errors on star imports?

jiayiyaij:
@amyreese thanks for the advice, I didn't know we could do that. Updated with using `QualifiedNameProvider` which is sufficient for our case

amyreese:
thank you!

Issue #313, opened 2023-04-12
no_assert_equals is incomplete


cricalix:
The rule `no_assert_equals` is incomplete; it only checks for `assertEquals`. There are probably 7 cases that the linter could fix, according to the Python 3 docs.

I'm happy to produce a PR for this, but before I do (and have to do it a different way):
* One rule file per deprecated alias, with one match and replace per file?
* One rule file for all deprecated aliases with a loop to match and replace?

amyreese:
I think a single rule/file that handles all deprecated aliases would be better, and result in less duplicate code and fewer visits to the syntax tree.

PR #314, opened 2023-04-13
Add more deprecated unittest asserts to lint/fix

cricalix:
## Summary

The existing `no_asserts_equals` only deals with a single case of [deprecated unittest TestCase method aliases](https://docs.python.org/3/library/unittest.html#deprecated-aliases). Since there are at least 7 that are easily handled, this change defines all 7, tests for them, and does the relevant replacements.

Python 2 does not define some of these entries as deprecated, so the ability to tie rules (and even "sub rules"?) to versions would be nice - though perhaps no one uses Fixit with Python 2.

For a reason I cannot fathom, git has lost the fact I did a `git mv` then replaced the content, so the history is lost.

Fixes issue #313.

This change potentially works for other test frameworks if they exposed the same aliases.

## Test Plan

`hatch run test`


amyreese:
Thank you!

Issue #363, opened 2023-07-21
Broken docs


arembridge:
https://github.com/Instagram/Fixit/blob/cf302f483edd285fcedfbdf6443f8b8bdd2d5006/docs/guide/quickstart.rst?plain=1#L164

I tried following the quickstart docs but the code has an error

```python
if name.value in ('"Paul"', "'Paul'"):
```

should be 

```python
if node.value in ('"Paul"', "'Paul'"):
```

PR #364, opened 2023-07-21
Fix broken docs

arembridge:
## Summary

Docs had some broken code that would not run (and didn't make sense).  Relates to issue: https://github.com/Instagram/Fixit/issues/363

## Test Plan

n/a


arembridge:
Tried signing the CLA, but it doesn't seem to have worked?...

amyreese:
Thank you!

Issue #365, opened 2023-07-24
Option for excluding files


slafs:
Hello! First of all, congrats on the v2 release. Really excited about the project!

Is there currently any way to exclude certain files from being processed?
We have a bunch of auto-generated Alembic migration files,
that we have no intention of fixing/linting.

The closest I saw to this feature is #354,
but I wouldn't want to edit those files in any way
(i.e. add a marker on top).

Instead I'd prefer to pass a glob string as an option
(either in CLI or in config file -
e.g. `--exclude '**/alembic/versions/**'`)
to either `lint` or `fix` commands.

amyreese:
As a CLI flag, this is simple: we can pass an `exclude` kwarg to `trailrunner.walk(...)` to apply excludes when walking. This would be a relatively straightforward PR adding an appropriate `exclude_paths` value to the `Options` dataclass and then passing that to trailrunner.

Providing this from a config file will be a bit more tricky, since we would need to materialize configs for each candidate path before we can exclude them, and that normally only happens in `fixit_file` after we already have spawned child processes. It might be useful instead to move towards materializing configs in `fixit_paths` based on the parent directory, use that when walking to exclude paths, and then pass those configs through to `fixit_file`.

This would also have the benefit of frontloading disk access, allow more caching of config objects, and give us access to the materialized set of lint rules for each file in the main process, which would enable work for #321 to preform warmup before spawning child processes.

datur:
It would also be really cool if the linter could skip over a file if there are no rules enable for a given file. Currently it seems all files are visited regardless of if a rule is enabled or not!

ggueret:
We’re eagerly awaiting this feature, so I’d love to help push it forward! Would you be open to a PR for the CLI flag implementation? I could help with adding the `exclude_paths` value to the `Options` dataclass and passing it to `trailrunner`. For the configuration file part, I understand it’s a bit more complex, but I’m happy to discuss and contribute if needed.

MakonnenMak:
+1 being able to define this in the `toml` file similar to ruff and black would be great as well.

PR #379, opened 2023-08-15
Skip files if no rules are enabled

datur:
## Summary
I had suggested on issue https://github.com/Instagram/Fixit/issues/365 that it would be nice to automatically skip a file if no rules were enabled for a given file https://github.com/Instagram/Fixit/issues/365#issuecomment-1678542798.

To avoid unnecessary processing of files if no rules are enabled for a given file the config is now checked before processing a file and if no rules are enabled the file is skipped.

## Test Plan
Two smoke tests have been added to confirm files are skipped if no rules are enabled.

### DIFF
--- a/src/fixit/api.py
+++ b/src/fixit/api.py
@@ -73,7 +73,8 @@ def fixit_bytes(
     Lint raw bytes content representing a single path, using the given configuration.
 
     Yields :class:`Result` objects for each lint error or exception found, or a single
-    empty result if the file is clean.
+    empty result if the file is clean. A file is considered clean if no lint errors or
+    no rules are enabled for the given path.
     Returns the final :class:`FileContent` including any fixes applied.
 
     Use :func:`capture` to more easily capture return value after iterating through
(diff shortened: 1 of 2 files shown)


amyreese (changes requested):
There are two main issues with this approach:

1) by skipping the call to `fixit_bytes()`, we are not generating a clean `Result()` object for the file, which will result in the CLI output under-counting the number of clean file results.

2) this only catching cases where the *file config* has no enabled rules/modules, but doesn't prevent visiting files when the number of rules actually discovered at runtime is empty (eg, passing `--tags` to CLI or combination of enable/disable that results in no selected rules).

I think the better solution is to modify `fixit_bytes()` instead such that it generates a clean result and aborts early if the result of `collect_rules(...)` is empty.

datur:
Thanks for the review @amyreese I believe I have addressed your change requests now! Let me know if you have any other feedback 😄

amyreese (approved):

amyreese:
Thank you!

Issue #387, opened 2023-09-06
feature: implement LSP


llllvvuu:
A bit of a long shot, but [`ruff` implements it](https://github.com/astral-sh/ruff-lsp). As do [ESLint](https://github.com/microsoft/vscode-eslint) and [Biome f.k.a. Rome](https://github.com/biomejs/biome/tree/main/crates/rome_lsp) for JavaScript.

Benefit of a language server:
* [perf](http://mitchgordon.me/software/2021/06/28/why-vscode-eslint-fast.html): persistent process throughout an editing session
* code actions: treat single issues (ignore <rule> for this line, ignore <rule> for this file, fix this line)

[LSP docs](https://microsoft.github.io/language-server-protocol/)
[`pygls` docs](https://github.com/openlawlibrary/pygls)

Potentially lower-hanging fruits:

- [x] Accept code input from stdin: https://github.com/Instagram/Fixit/pull/388
- [ ] Make it daemonizable: [example](https://mypy.readthedocs.io/en/latest/mypy_daemon.html). Probably not that worth it compared to just implementing LSP

Related: https://github.com/Instagram/Fixit/issues/122  - many VSCode extensions are built using LSP. Just wanted to flag the protocol since there is a failure mode of building a VSCode extension without the protocol and thus not supporting other editors.

Workaround: let third-parties build daemons/LSPs off of the [Fixit API](https://fixit.readthedocs.io/en/latest/api.html)

Issue #122, opened 2020-09-07
Integrate Fixit into a VSCode plugin


jimmylai:
Integrate Fixit into a VSCode plugin (and potentially any other IDE), so developer can easily install it and use it on VSCode.

Fixit can be run automatically when developers save source file updates and show warning inline with wavy underline or lightbulb UIs.

Then developers can click on the UI  or press some shortcut to apply autofix.

Some helper is provided to make this easy to implement, e.g. `lint_file` returns a list of violation reports (BaseLintRuleReport) and provides `patch` property when autofix is available.

https://github.com/Instagram/Fixit/blob/master/fixit/rule_lint_engine.py#L54

llllvvuu:
Here's a rather janky way to get VSCode diagnostics on load/save: https://github.com/llllvvuu/efm-tool-definitions.yaml/blob/cb90f12516a27a8e1820600afe94229c04bb5c19/tool-definitions.yaml#L42-L55

The above becomes lint-as-you-type if https://github.com/Instagram/Fixit/pull/388 is merged.

That leaves the remaining task of supporting Code Actions, which requires implementing LSP: #387

PR #388, opened 2023-09-06
feat: support linting stdin

llllvvuu:
Similarly for `fixit fix`. The API is [inspired by Prettier](https://prettier.io/docs/en/cli.html#--stdin-filepath).

## Summary
Related to #122 but does not fully resolve it. It supports a lint-as-you-type workflow, whereas only a lint-on-save workflow could be supported before. Full resolution of #122 would require #387.

## Test Plan
Extended smoke tests.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,7 +1,7 @@
 root = true
 
 [*.{py,pyi,toml,md}]
-charset = "utf-8"
+charset = utf-8
 end_of_line = lf
 indent_size = 4
 indent_style = space
--- a/docs/guide/commands.rst
+++ b/docs/guide/commands.rst
@@ -46,7 +46,9 @@ The following options are available for all commands:
 ``lint``
 ^^^^^^^^
 
-Lint one or more paths, and print a list of lint errors.
+Lint one or more paths, and print a list of lint errors. If "-" is given as the
+first path, then the second given path will be used for configuration lookup
+and error messages, and the input read from STDIN.
 
 .. code:: console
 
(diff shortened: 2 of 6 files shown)


llllvvuu:
CLA submitted

amyreese (changes requested):
I like this as functionality to support, but I'd rather push this functionality into the supported API, rather than handling stdin at the CLI layer.

Take a look at how we did this for µfmt, where passing `Path("-")` (what you get from click.Path(allow_dash=True) as the first element of `paths` to `ufmt_paths(...)` triggers handling for stdin.

Adding `fixit_stdin` to the API to handle reading stdin and wrap `fixit_bytes` would be the first step, then update `fixit_paths` to detect the stdin sentinel and defer to `fixit_stdin` instead of normal file processing.

Is that something you would be interested in tackling, or would you rather I adapt your work accordingly?

llllvvuu:
Just took a look at ufmt CLI and happy to make the changes. Would we still need `--stdin-filepath`? (for `generate_config`)

Like:
```sh
fixit lint - --stdin-filepath foo.py
```

amyreese:
> Just took a look at ufmt CLI and happy to make the changes. So to confirm, intended usage is
> 
> ```shell
> fixit lint - path/to/foo.py  # formats stdin, treating it as if it is located at path/to/foo.py
> ```

Correct :)

llllvvuu:
Just made the changes, thanks again for the ufmt reference!

I wasn't able to leave `cli.py` 100% unchanged since `fixit fix` needs to disable `--interactive` (as STDIN is occupied) and redirect `Result`s to STDERR (as STDOUT is occupied). But let me know what you think

amyreese:
Thank you!

Issue #358, opened 2023-07-15
Better story around lint rules adding imports


amyreese:
Adding imports with LibCST is a sticky subject because they are normal statements mixed inside `Module.body`. It's not currently possible to do a simple addition with LibCST alongside other changes because that would require replacing the entire root `Module` object with one that contains the new statements, nullifying any other transforms that happened during linting.

Consider offering a mechanism to include any desired additional imports with each reported violation, and then trigger AddImportsVisitor after applying any other autofixes, so that all imports are added at once, and can operate on the already-fixed module object.

Issue #352, opened 2023-07-12
Automatic #lint-fixme suppressions


amyreese:
Considerations:

- default to `# lint-fixme: <rule>`
- enable with `--add-fixme` flag?
- only for lints without fixes?
- only with `fix` command?

samueljsb:
We needed a tool to do something like this (to mass-ignore existing violations when we added a new rule without an obvious auto-fix) so @meshy and I wrote a separate tool to do it: https://pypi.org/project/silence-fixit-error/

PR #390, opened 2023-09-07
feat: add `fixit lsp` subcommand

### LINKED ISSUE #387: feature: implement LSP

llllvvuu:
resolves #387 #122

Note: easier to read in unified view since it's new code.
Note: uses the popular [pygls](https://github.com/openlawlibrary/pygls) ([Apache 2.0](https://github.com/openlawlibrary/pygls/blob/main/LICENSE.txt)) implementation of the protocol.

https://github.com/Instagram/Fixit/assets/5601392/2dcd7447-431e-49a9-a383-aa424d4fbc6b

## Summary

Support for:
- [x] `textDocument/didOpen`, `textDocument/didChange` -> `textDocument/publishDiagnostics`
- [x] `textDocument/formatting`

Not supported yet

- [ ] `textDocument/codeAction`, `workspace/executeCommand` (possibly blocked on #352 #358)
- [ ] `workspace/didChangeWatchedFiles` to invalidate the config cache
- [ ] VSCode frontend (documented [Generic LSP Client](https://github.com/llllvvuu/vscode-glspc) for now but probably want Fixit-specific extension/branding)

If this PR gets merged I will create follow-up issues for these items.

## Test Plan

Added new smoke test for the new `fixit lsp` subcommand.

### DIFF
--- a/src/fixit/cli.py
+++ b/src/fixit/cli.py
@@ -220,7 +220,7 @@ def lsp(
     tcp: Optional[int],
     ws: Optional[int],
     debounce_interval: float,
-):
+) -> None:
     """
     Start server for:
     https://microsoft.github.io/language-server-protocol/
--- a/src/fixit/cli.py
+++ b/src/fixit/cli.py
@@ -15,7 +15,7 @@ from fixit import __version__
 
 from .api import fixit_paths, print_result
 from .config import collect_rules, generate_config, parse_rule
-from .ftypes import Config, Options, QualifiedRule, Tags
+from .ftypes import Config, LSPOptions, Options, QualifiedRule, Tags
 from .rule import LintRule
 from .testing import generate_lint_rule_test_cases
 from .util import capture
@@ -201,6 +201,40 @@ def fix(
 
     splash(visited, dirty, autofixes, fixed)
     ctx.exit(exit_code)
+
+
+@main.command()
+@click.pass_context
+@click.option("--stdio", type=bool, default=True, help="Serve LSP over stdio")
+@click.option("--tcp", type=int, help="Port to serve LSP over")
+@click.option("--ws", type=int, help="Port to serve WS over")
+@click.option(
+    "--debounce-interval",
+    type=float,
+    default=LSPOptions.debounce_interval,
+    help="Delay in seconds for server-side debounce",
+)
+def lsp(
+    ctx: click.Context,
+    stdio: bool,
+    tcp: Optional[int],
+    ws: Optional[int],
+    debounce_interval: float,
+):
+    """
+    Start server for:
+    https://microsoft.github.io/language-server-protocol/
+    """
+    from .lsp import LSP
+
+    main_options = ctx.obj
+    lsp_options = LSPOptions(
+        tcp=tcp,
+        ws=ws,
+        stdio=stdio,
+        debounce_interval=debounce_interval,
+    )
+    LSP(main_options, lsp_options).start()
 
 
 @main.command()
--- a/src/fixit/ftypes.py
+++ b/src/fixit/ftypes.py
@@ -184,6 +184,18 @@ class Options:
 
 
 @dataclass
+class LSPOptions:
+    """
+    Command-line options to affect LSP runtime behavior
+    """
+
+    tcp: Optional[int]
+    ws: Optional[int]
+    stdio: bool = True
+    debounce_interval: float = 0.5
+
+
+@dataclass
 class Config:
     """
     Materialized configuration valid for processing a single file.
--- a/src/fixit/lsp.py
+++ b/src/fixit/lsp.py
@@ -9,7 +9,6 @@ from pathlib import Path
 from typing import Any, Callable, cast, Dict, Generator, List, Optional, TypeVar
 
 import pygls.uris as Uri
-
 from lsprotocol.types import (
     Diagnostic,
     DiagnosticSeverity,
@@ -24,13 +23,13 @@ from lsprotocol.types import (
     TextEdit,
 )
 from pygls.server import LanguageServer
+from pygls.workspace.text_document import TextDocument
 
-from fixit import __version__
-from fixit.util import capture
-
+from .__version__ import __version__
 from .api import fixit_bytes
 from .config import generate_config
 from .ftypes import Config, FileContent, LSPOptions, Options, Result
+from .util import capture
 
 
 class LSP:
@@ -65,7 +64,7 @@ class LSP:
         return self._config_cache[path]
 
     def diagnostic_generator(
-        self, uri: str, autofix=False
+        self, uri: str, autofix: bool = False
     ) -> Optional[Generator[Result, bool, Optional[FileContent]]]:
         """
         LSP wrapper (provides document state from `pygls`) for `fixit_bytes`.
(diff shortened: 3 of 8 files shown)


amyreese:
Sorry for the delay. I'll try to review this and your other PR's some time next week.

src/fixit/ftypes.py
```diff
@@ -182,6 +182,18 @@ class Options:
...
+@dataclass
+class LspOptions:
+    """
+    Command-line options to affect LSP runtime behavior
+    """
+
+    tcp: Optional[int]
+    ws: Optional[int]
+    stdio: bool = True
+    debounce_interval: float = 0
```
amyreese, line 194: Is it unreasonable to set a default interval of 0.5 or 1 second?

src/fixit/cli.py
```diff
@@ -194,6 +205,38 @@ def fix(
...
 
+@main.command()
+@click.pass_context
+@click.option("--stdio", type=bool, default=True, help="Serve LSP over stdio")
+@click.option("--tcp", type=int, help="Port to serve LSP over")
+@click.option("--ws", type=int, help="Port to serve WS over")
+@click.option(
+    "--debounce-interval",
+    type=float,
+    default=0,
```
amyreese, line 216: Reference `LSPOptions.debounce_interval` so that the CLI shares the same default value?

src/fixit/util.py
```diff
@@ -87,3 +88,36 @@ def append_sys_path(path: Path) -> Generator[None, None, None]:
...
+    The decorated function will return None immediately,
+    ignoring the delayed return value of `f`.
+    """
+
+    def decorator(f: VoidFunction) -> VoidFunction:
+        if interval <= 0:
+            return f
+        return cast(VoidFunction, Debouncer(f, interval))
+
+    return decorator
```
amyreese, line 123: I should add some docs to this module to clarify, but the `util` module is meant more for additional helpers for use by clients/frontends. Is the debouncing helper something you expect would be used by something other than the LSP module? Maybe just put it next to the LSP class otherwise?

src/fixit/lsp.py
```diff
@@ -0,0 +1,162 @@
...
+        self.lsp.publish_diagnostics(uri, diagnostics, version=version)
+
+    def debounced_validator(self, uri: str) -> Callable[[int], None]:
+        """
+        Per-URI debounced validation function. See: LSP.validate
+        """
+        if uri not in self._debounced_validators:
+            self._debounced_validators[uri] = debounce(
+                self.lsp_options.debounce_interval
+            )(lambda version: self.validate(uri, version))
```
amyreese, line 109: Would prefer named functions or partials over lambdas when possible.

src/fixit/lsp.py
```diff
@@ -0,0 +1,162 @@
...
+        """
+        Side-effect: mutates self.lsp to make it respond to
+        `textDocument/didOpen`, `textDocument/didChange`, `textDocument/formatting`
+        """
+
+        @self.lsp.feature(TEXT_DOCUMENT_DID_OPEN)
+        def _(params: DidOpenTextDocumentParams):
+            self.debounced_validator(params.text_document.uri)(
+                params.text_document.version
+            )
```
amyreese, line 122: I'm not super fond of the "call a method to get a cached callable and then call that" pattern.  Would prefer if this could be just `self.validate(uri, version)` and let the `validate` method handle debouncing as necessary.

src/fixit/lsp.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+class LSP:
+    """
+    Server for the Language Server Protocol.
+    Provides diagnostics as you type, and exposes a formatter.
+    https://microsoft.github.io/language-server-protocol/
+    """
+
+    def __init__(self, fixit_options: Options, lsp_options: LspOptions):
```
amyreese, line 41: A number of methods here need type annotations.

src/fixit/lsp.py
```diff
@@ -0,0 +1,162 @@
...
+            if not formatted_content:
+                return None
+
+            doc = self.lsp.workspace.get_document(params.text_document.uri)
+            entire_range = Range(
+                start=Position(line=0, character=0),
+                end=Position(line=len(doc.lines) - 1, character=len(doc.lines[-1])),
+            )
+
+            return [TextEdit(new_text=formatted_content.decode(), range=entire_range)]
```
amyreese, line 151: Would like this to be in a separate method the way `validate` is. Maybe just something like `format(...) -> List[TextEdit]`?

src/fixit/lsp.py
```diff
@@ -0,0 +1,162 @@
...
+        return self._debounced_validators[uri]
+
+    def set_handlers(self) -> None:
+        """
+        Side-effect: mutates self.lsp to make it respond to
+        `textDocument/didOpen`, `textDocument/didChange`, `textDocument/formatting`
+        """
+
+        @self.lsp.feature(TEXT_DOCUMENT_DID_OPEN)
+        def _(params: DidOpenTextDocumentParams):
```
amyreese, line 119: Would prefer these methods to be named if possible.

amyreese (changes requested):
Sorry for the delay in reviewing this, and I really appreciate your time in putting this together. My only experience with writing an LSP is using the vscode templates, so forgive me if some of this goes against common patterns. I'm mostly interested in keeping the code more readable and maintainable, and I had a bit of difficulty following what was happening here. 😅

I don't want to impose on your time too much, so just let me know if you'd rather me be the one to adopt these suggestions, and I'll make sure to keep you as the primary author on the PR.

Cheers!

llllvvuu:
No worries, might be delayed as I don't have Wi-fi this month, but I'm happy to make the changes as they all seem sensible!

src/fixit/lsp.py
llllvvuu: Added return types

llllvvuu:
Updated, sorry for the delay!

povilasb:
This is cool!

Hope it will get merged 🤞

amyreese (approved):

Issue #375, opened 2023-08-08
Circular fix applied involving Path and ComparePrimitivesByEqual


nth10sd:
Create a `testcase.py` file:

```
from pathlib import Path
from tempfile import TemporaryDirectory

with TemporaryDirectory() as f:
    fn_ = Path(f) / "test.txt"
    fn_.write_text("Testing!\n")

    assert fn_.read_text() == "Testing!\n"  # No fixit error

    assert (Path(f) / "test.txt").samefile(fn_)  # Assert passes, so they are identical
    assert (Path(f) / "test.txt").read_text() == "Testing!\n"  # fixit shows an error
```

Run the following command once:

```
grep "shows an error" testcase.py ; fixit lint --diff testcase.py ; fixit fix --automatic testcase.py ; grep "shows an error" testcase.py
```

You will get the first output:

```
    assert (Path(f) / "test.txt").read_text() == "Testing!\n"  # fixit shows an error
testcase.py@11:11 CompareSingletonPrimitivesByIs: Comparisons to singleton primitives should not be done with == or !=, as they check equality rather than identiy. Use `is` or `is not` instead. (has autofix)
--- a/testcase.py
+++ b/testcase.py
@@ -10,2 +10,2 @@
     assert (Path(f) / "test.txt").samefile(fn_)  # Assert passes, so they are identical
-    assert (Path(f) / "test.txt").read_text() == "Testing!\n"  # fixit shows an error
+    assert (Path(f) / "test.txt").read_text() is "Testing!\n"  # fixit shows an error
🛠️  1 file checked, 1 file with errors, 1 auto-fix available 🛠️
testcase.py@11:11 CompareSingletonPrimitivesByIs: Comparisons to singleton primitives should not be done with == or !=, as they check equality rather than identiy. Use `is` or `is not` instead. (has autofix)
🛠️  1 file checked, 1 file with errors, 1 auto-fix available, 1 fix applied 🛠️
    assert (Path(f) / "test.txt").read_text() is "Testing!\n"  # fixit shows an error
```

Run the command again, you will get this second set of fixes back to the original:

```
    assert (Path(f) / "test.txt").read_text() is "Testing!\n"  # fixit shows an error
testcase.py@11:11 ComparePrimitivesByEqual: Don't use `is` or `is not` to compare primitives, as they compare references. Use == or != instead. (has autofix)
--- a/testcase.py
+++ b/testcase.py
@@ -10,2 +10,2 @@
[...]

PR #391, opened 2023-09-07
fix(compare_singleton_primitives_by_is): compare Name directly instead of QualifiedName

### LINKED ISSUE #378: Incorrect suggestion to fix string comparison
It seems this autofix is changing the logic for no good reason?

```
users/account.py@370:15 CompareSingletonPrimitivesByIs: Comparisons to singleton primitives should not be done with == or !=, as they check equality rather than identiy. Use `is` or `is not` instead. (has autofix)
--- a/account.py
+++ b/account.py
@@ -369,3 +369,3 @@
         if form.cleaned_data["email"]:
-            if form.cleaned_data["email"].lower() != (request.user.email or "").lower():
+            if form.cleaned_data["email"].lower() is not (request.user.email or "").lower():
                 if User.objects.filter(
```

```
fixit -V
fixit, version 2.0.0.post1
```

llllvvuu:
## Summary

fixes #378 and #375 which are caused by https://github.com/Instagram/LibCST/issues/389.

Qualifying a name will never make something that wasn't already True/False/None into a True/False/None, so it's not necessary.

## Test Plan

Added two Valid cases, one of which breaks on HEAD^.

### DIFF
--- a/src/fixit/rules/compare_singleton_primitives_by_is.py
+++ b/src/fixit/rules/compare_singleton_primitives_by_is.py
@@ -3,7 +3,7 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import FrozenSet, Union
+from typing import cast, FrozenSet, Union
 
 import libcst as cst
 from libcst.metadata import QualifiedName, QualifiedNameProvider, QualifiedNameSource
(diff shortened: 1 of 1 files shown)


amyreese:
Thank you!

I modified the fix slightly, so that it still checks for `cst.Name`, but also still uses the existing set of `QualifiedName` values.

Issue #405, opened 2023-11-02
`lint-fixme` comments ignored inside list comprehensions


samueljsb:
`lint-fixme` comments on nodes inside comprehensions are not respected.

In the following file (`t.py`), we would expect the violation to be reported for the first function, but not for either of the other two.

```python
def f(l):
    return [
        i for i in l
        # lint-fixme: CollapseIsinstanceChecks
        if isinstance(i, int) or isinstance(i, float)
    ]


def g(l):
    return [
        i for i in l
        # lint-fixme: CollapseIsinstanceChecks
        if isinstance(i, int) or isinstance(i, float)
    ]


def h(l):
    return [
        i for i in l
        if isinstance(i, int) or isinstance(i, float)  # lint-fixme: CollapseIsinstanceChecks
    ]
```

However, the violation is reported for all three:

```console
$ fixit lint t.py
t.py@5:11 CollapseIsinstanceChecks: Multiple isinstance calls with the same target but different types can be collapsed into a single call with a tuple of types. (has autofix)
t.py@13:11 CollapseIsinstanceChecks: Multiple isinstance calls with the same target but different types can be collapsed into a single call with a tuple of types. (has autofix)
t.py@20:11 CollapseIsinstanceChecks: Multiple isinstance calls with the same target but different types can be collapsed into a single call with a tuple of types. (has autofix)
🛠️  1 file checked, 1 file with errors, 3 auto-fixes available 🛠️
```

```console
$ python --version
Python 3.11.5

$ fixit --version
fixit, version 2.1.0
```

samueljsb:
I've found another one that's causing us problems: decorators.

```python
from dataclasses import dataclass


@dataclass()
class C:
    pass


# lint-fixme: ExplicitFrozenDataclass
@dataclass()
class C:
    pass


@dataclass()  # lint-fixme: ExplicitFrozenDataclass
class C:
    pass
```

We'd expect the first function to be reported, and the other two to be silenced.

```console
t.py@4:0 ExplicitFrozenDataclass: When using dataclasses, explicitly specify a frozen keyword argument. Example: `@dataclass(frozen=True)` or `@dataclass(frozen=False)`. Docs: https://docs.python.org/3/library/dataclasses.html (has autofix)
[...]

PR #413, opened 2023-12-01
Demonstrate fixme comments that are not being respected

samueljsb:
## Test Plan

Failing tests have been added to demonstrate (most of) the bugs reported in #405. I have not been able to add one for the function definition problem because there are cases in this test that need to define functions without errors, so I can't just report `functionDef`.


amyreese:
Thank you so much for these examples! I've got a fix in #441, and squashed this PR with some additional test cases into #442 (retaining your authorship).

PR #437, opened 2024-04-06
make format of fixit's terminal output configurable

### LINKED ISSUE #305: Alternate output formats
- Something JSON
- User-customizable maybe?

### LINKED ISSUE #397: Conventional file path/line number separator (`:` instead of `@`)
The output when a line is reported uses the `@` character to separate the filename from the line number, e.g:

```
path/to/some/module.py@91:12 An error is at this location.
```

This differs from most other tools, which use a colon. The colon syntax is understood by many other tools that parse output (e.g. [`mypy-json-report`](https://pypi.org/project/mypy-json-report/)) and the file finder in most editors (e.g. VS Code). e.g:

```
path/to/some/module.py:91:12 An error is at this location.
```

Would the maintainers be open to changing the output format to be the same as other tools? It would (imo) make fixit output much easier to work with.

jvllmr:
## Summary
This PR makes the format of Fixit's output configurable.
I hope I got everything right 😸 

fixes #397 
closes #305

### DIFF
--- a/docs/guide/commands.rst
+++ b/docs/guide/commands.rst
@@ -43,6 +43,16 @@ The following options are available for all commands:
     two tags.
 
 
+.. attribute:: --output-format / -o FORMAT_TYPE
+
+    Override how Fixit prints violations to the terminal.
+
+    See :attr:`output-format` for available formats.
+
+.. attribute:: --output-template TEMPLATE
+
+    Override the python formatting template to use with ``output-format = 'custom'``.
+
 ``lint``
 ^^^^^^^^
 
@@ -72,7 +82,7 @@ the input read from STDIN, and the fixed output printed to STDOUT (ignoring
     $ fixit fix [--interactive | --automatic [--diff]] [PATH ...]
 
 .. attribute:: --interactive / -i
-    
+
     Interactively prompt the user to apply or decline suggested fixes for
     each auto-fix available. *default*
 
--- a/docs/guide/configuration.rst
+++ b/docs/guide/configuration.rst
@@ -165,6 +165,13 @@ The main configuration table.
     - ``fixit``: Fixit's default output format.
     - ``vscode``: A format that provides clickable paths for Visual Studio Code.
 
+    .. note::
+
+        The default output format is planned to change to ``vscode`` in
+        the next feature release, expected as part of ``v2.3`` or ``v3.0``.
+        If you are sensitive to output formats changing, specify your preferred
+        format in your project configs accordingly.
+
 .. attribute:: output-template
     :type: str
 
--- a/docs/guide/configuration.rst
+++ b/docs/guide/configuration.rst
@@ -152,40 +152,43 @@ The main configuration table.
     :class:`~fixit.Formatter` interface.
 
 .. attribute:: output-format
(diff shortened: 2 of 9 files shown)


amyreese:
First off, I greatly appreciate the PR, thank you! That said, I have some concerns with the current design.

I'm not fond of passing the original config object back for every single result, for a few reasons:

- that's a lot of extra data to pickle/unpickle, especially when linting a large number of files in parallel
- that opens the door to providing inconsistent formatting across multiple results in a single run, in cases where a subdirectory's config specifies an alternative format
- that makes it more complicated to override output format with a matching `--output-format` CLI flag

Instead, I think what would make most sense is to pull the config for the current working directory in the CLI module, and then pass the relevant output format straight to `print_result`. This would ensure a consistent formatting style for the run based on where Fixit is run from. This would also make it relatively trivial to implement `--output-format` and use that instead of the config when given.

---

On the actual config mechanism, I think rather than enabling this as a free-form text field (which I suspect would be overly complicated for 99% of use cases), a set of "named" formats would be better.  Eg, `output-format = "vscode"` provides clear intent, and is likely less error-prone than copy-pasting a bunch of raw format strings.

A standardized JSON result could be defined with `output-format = "json"`, and the proposed free-form behavior could then be implemented as something like `output-format = "custom"` and `output-format-template = "..."`, or maybe just `output-format = "template:<template string>".

---

Given the popularity of VS Code and the utility of the vscode-compatible format, perhaps it would be worth discussing a plan to make that the default format style in a future release. Ie, implement the selectable output format and maintain the current default, with a "vscode" format style, and set a target for a future date or release version where the "vscode" style will become the default. 

I suspect that will make more folks happy in the long run.

---

I'm happy to make most of these proposed design changes if you'd prefer to not spend more time on it. Let me know, and I'll be sure to preserve author credit for you on any changes I make.

jvllmr:
Thanks for the very detailed feedback and I see your point.

I'd like to try integrating your suggestions myself.
I'm not quite sure how the JSON format should look like yet, but I'm sure we'll figure it out 😄 

I'll probably find some time to work on this next weekend.

amyreese:
I don't think the JSON format needs to be included in this PR, but mentioned it more for context in the overall design direction. Happy to worry about a JSON format in a different PR or later in the future. Cheers!

jvllmr:
I adjusted the PR according to your suggestion. There surely is room for improvement. Especially for the tests. But I first want to check with you if the general idea is correct.

duzumaki:
hey @amyreese any thoughts on his pr? would be great to have colon instead of @ so it works better with code editors when jumping to lines

amyreese:
Sorry, just been trying to find the time to do a full review on the new revisions. I will try to get to it soon.

amyreese (approved):
- Reworked the data types to just use an enum, and to deduplicate the config handling and reuse the existing loader.
- Added a handful of extra test cases.
- Updated and reworded docs, added note on future plans.

Overall very happy with where this is now. Added a TODO to implement the JSON output mode at some point in the future.

amyreese:
@jvllmr thank you for your help on this feature. It is very much appreciated and will go out with the next feature release (hopefully soon™). 💜

PR #450, opened 2024-05-01
Extract node_comments to new module, return nodes instead of strings

amyreese:
* (to be filled)


src/fixit/engine.py
```diff
@@ -54,6 +54,7 @@ def __init__(self, path: Path, source: FileContent) -> None:
         self.source = source
         self.module: Module = parse_module(source)
         self.timings: Timings = defaultdict(lambda: 0)
+        self.wrapper = MetadataWrapper(self.module)
```
zsol, line 57: nit: you can always get back to the module with `self.wrapper.module`

zsol (approved):

PR #451, opened 2024-05-01
Add helpers to find and add suppressions comments to nodes

amyreese:
* (to be filled)

### DIFF
--- a/src/fixit/ftypes.py
+++ b/src/fixit/ftypes.py
@@ -19,6 +19,7 @@ from typing import (
     List,
     Optional,
     Sequence,
+    Set,
     Tuple,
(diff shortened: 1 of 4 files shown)


src/fixit/ftypes.py
```diff
@@ -29,6 +30,7 @@
 from libcst._add_slots import add_slots
 from libcst.metadata import CodePosition as CodePosition, CodeRange as CodeRange
 from packaging.version import Version
+from typing_extensions import Self
```
zsol, line 33:
```suggestion
from typing_extensions import Self  # available under `typing` from 3.11
```

PR #459, opened 2024-05-21
Pre-generate configs for the tree of files

jarshwah:
Experiments with pre-generating all config files in a top-down manner and sharing that with each sub-process, so that we don't have to re-generate a config-per-file.

### DIFF
--- a/src/fixit/api.py
+++ b/src/fixit/api.py
@@ -5,19 +5,28 @@
 
 import logging
 import sys
+import time
 import traceback
 from functools import partial
 from pathlib import Path
-from typing import Generator, Iterable, List, Optional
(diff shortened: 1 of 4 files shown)


src/fixit/api.py
```diff
@@ -186,12 +199,25 @@ def fixit_file(
 
 
 def _fixit_file_wrapper(
-    path: Path, *, autofix: bool = False, options: Optional[Options] = None
+    path: Path,
+    *,
+    autofix: bool = False,
+    options: Optional[Options] = None,
+    trees: Optional[Dict[Path, ConfigTree]] = None,
```
jarshwah, line 206: TODO: initialise the process pool with the config trees rather than serializing them into each task.

PR #460, opened 2024-05-31
Lintrule collections abc

surge119:
## Summary
Since python 3.3, the Collections Abstract Base Classes (ABC) has been moved to the `collections.abc` module. This lint rule checks and fixes (some) ABC imports from `collections` to `collections.abc`.

## Test Plan
The lint rule has several built in tests that were run using `hatch run test`

### DIFF
--- a/src/fixit/rules/deprecated_abc_import.py
+++ b/src/fixit/rules/deprecated_abc_import.py
@@ -88,6 +88,10 @@ class DeprecatedABCImport(LintRule):
             expected_replacement="import collections.abc.Container",
         ),
         Invalid(
+            "import collections.Container as cont",
+            expected_replacement="import collections.abc.Container as cont",
+        ),
+        Invalid(
             "from collections import defaultdict, Container",
             expected_replacement="from collections import defaultdict\nfrom collections.abc import Container",
         ),
--- a/src/fixit/rules/deprecated_abc_import.py
+++ b/src/fixit/rules/deprecated_abc_import.py
@@ -52,12 +52,12 @@ ABCS = frozenset(
 
 class DeprecatedABCImport(LintRule):
     """
-    Checks for the use of the deprecated collections ABC import. Since python 3.3, the Collections Abstract Base Classes (ABC) have been moved to `collections.abc`.
-    This `LintRule` checks that all ABC imports are under `collections.abc`.
+    Checks for the use of the deprecated collections ABC import. Since python 3.3,
+    the Collections Abstract Base Classes (ABC) have been moved to `collections.abc`.
+    These ABCs are import errors starting in Python 3.10.
     """
 
-    TASKS = {"safe"}
-    MESSAGE = "Since python 3.3, the Collections Abstract Base Classes (ABC) have been moved to `collections.abc`. This was a deprecation warning up until 3.9, and is an import error in 3.10."
+    MESSAGE = "ABCs must be imported from collections.abc"
     PYTHON_VERSION = ">= 3.3"
 
     VALID = [
--- a/src/fixit/rules/deprecated_abc_import.py
+++ b/src/fixit/rules/deprecated_abc_import.py
@@ -8,7 +8,6 @@ from typing import List, Optional, Union
 import libcst as cst
 
 import libcst.matchers as m
-from libcst._nodes.expression import Attribute, Name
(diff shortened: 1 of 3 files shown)


src/fixit/rules/deprecated_abc_import.py
```diff
@@ -0,0 +1,250 @@
...
+                    else []
+                )
+                if (
+                    isinstance(statement, ImportFrom)
+                    and statement.module
+                    and statement.module.value == "collections"
+                    # Ensure the imports in this statement are the ones we are looking for
+                    and set(import_names).intersection(set(self.imports_names))
+                ):
+                    return statement
```
amyreese, line 169: It should be possible to express most of this with a libcst matcher and reduce the number of instance checks, etc. Something like:

```
import libcst.matchers as m

if m.matches(node, m.SimpleStatementLine(body=m.ZeroOrMore(m.ImportFrom(...)))):
    ...
```

Something like `m.findall(...)` may also be helpful.

https://libcst.readthedocs.io/en/latest/matchers_tutorial.html

https://libcst.readthedocs.io/en/latest/matchers.html

src/fixit/rules/deprecated_abc_import.py
```diff
@@ -0,0 +1,250 @@
...
+            # Replacing the case where there are ABCs mixed with non-ABCs requires
+            # splitting a single import statement into two separate imports. This
+            # cannot be achieved in this method and is offloaded to leaving the module.
+            if not all(import_names_in_abc):
+                # We set this variable which triggers the `self.report` to be called
+                # in `leave_Module`. We report in the `leave_Module`
+                # so that we can add an additional `SimpleStatementLine` for the new
+                # import
+                self.update_module = True
+                self.imports_names = import_names
```
amyreese, line 131: initialize this in `__init__` with an appropriate type annotation?

src/fixit/rules/deprecated_abc_import.py
```diff
@@ -0,0 +1,250 @@
...
+                        ),
+                    )
+
+            self.report(node, replacement=node.with_changes(body=node_body))
+
+    def visit_ImportAlias(self, node: cst.ImportAlias) -> None:
+        """
+        This catches the `import collections.<ABC>` cases.
+        """
+        if (
```
amyreese, line 233: Not sure if it would be more clear to express this big conditional using a matcher decorator: https://libcst.readthedocs.io/en/latest/matchers_tutorial.html#Matcher-Decorators

surge119: Updated to use matcher in the conditional. Couldn't get decorators to function properly,

src/fixit/rules/deprecated_abc_import.py
```diff
@@ -0,0 +1,331 @@
...
+        ):
+            imp = m.findall(
+                node,
+                m.ImportFrom(
+                    module=m.Name("collections"),
+                    names=m.OneOf(
+                        [m.ImportAlias(name=m.Name(n)) for n in self.imports_names]
+                    ),
+                ),
+            )[0]
```
amyreese, line 193: Do you need the conditional `m.matches()` guarding this? Maybe just `m.findall()` and have a conditional on whether it finds something?

src/fixit/rules/deprecated_abc_import.py
```diff
@@ -0,0 +1,331 @@
...
+                    name=m.Attribute(
+                        value=m.Name("collections"),
+                        attr=m.OneOf(*[m.Name(abc) for abc in ABCS]),
+                    )
+                ),
+            )
+            # This is necessary for the typecheck. Otherwise when replacing the node
+            # and using `attr=node.name.attr`, the type checker will complain since
+            # node.name is `Union[cst.Name | cst.Attribute]`
+            and isinstance(node.name, cst.Attribute)
```
amyreese, line 271: You can use `libcst.ensure_type(...)` to constrain the type at the time you want to use it for cases like this where the matcher has already selected nodes matching that expected type.

Something like `attr=cst.ensure_type(node.name, cst.Attribute).attr`

https://libcst.readthedocs.io/en/latest/best_practices.html#avoid-isinstance-when-traversing

amyreese (approved):

amyreese:
Thank you! 🥂

PR #474, opened 2024-07-25
Update DeprecatedABCImport Rule

surge119:
## Summary
In the instance of there being a try except that falls back to the deprecated import, this rule would attempt to change the fallback. An example:
```
try:
    from collections.abc import Container
except ImportError:
    from collections import Container
```
This above example should be valid, but was previously not. Previously, the lint rule would attempt to fix the import in the except block

## Test Plan
make

### DIFF
--- a/src/fixit/rules/deprecated_abc_import.py
+++ b/src/fixit/rules/deprecated_abc_import.py
@@ -6,10 +6,10 @@
 from typing import List, Optional, Union
 
 import libcst as cst
-
 import libcst.matchers as m
 
 from fixit import Invalid, LintRule, Valid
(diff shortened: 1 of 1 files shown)


amyreese (changes requested):
I think it would be better/faster to use `ParentNodeProvider` metadata to see if an import is directly nested inside a try/except block, rather than walking *every* try/except in the entire file regardless of whether it may or may not contain any imports. I think the code would also be simpler that way, and wouldn't require caching a set of ignored nodes.

amyreese (approved):
Less code 👍 👍

PR #475, opened 2024-07-30
Add More Metrics to be Logged by LoggerHook

surge119:
## Summary
Added the following metrics:
- Violation Count per Rule per File
- Violation Count per Rule per File with Autofixes
- Total Violation Count per File

## Test Plan
make

### DIFF
--- a/src/fixit/engine.py
+++ b/src/fixit/engine.py
@@ -115,7 +115,7 @@ class LintRunner:
 
                 yield violation
 
-        self.metrics["ViolationCount.Total"] = count
+        self.metrics["Count.Total"] = count
 
         if metrics_hook:
             metrics_hook(self.metrics)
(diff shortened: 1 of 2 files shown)


src/fixit/engine.py
```diff
@@ -103,14 +103,20 @@ def visit_hook(name: str) -> Iterator[None]:
         wrapper.visit_batched(rules)
         count = 0
         for rule in rules:
+            self.metrics[f"ViolationCount.{rule.name}"] = len(rule._violations)
+            self.metrics[f"ViolationCountWithReplacement.{rule.name}"] = 0
```
amyreese, line 107: What about just `Count` and `FixCount` for brevity? Esp in a potential world where auto-fixes aren't just strict "node replacements".

amyreese (approved):

PR #479, opened 2024-08-09
Improved the Handling of Errors in Rule Collection

surge119:
## Summary
If a rule defined in a config contains an error or causes an error in Fixit, the linter will raise an unrecoverable error. This PR adds error handling to catch and log those errors as warnings, and continue linting

## Test Plan
```
$ hatch run fixit lint ./src/fixit/api.py
WARNING:fixit.config:Failed to load rules 'fixit.rules': CollectionError: could not import rule(s) fixit.rules
🧼 1 file clean 🧼

# Run with good rules
$ hatch run fixit lint ./src/fixit/api.py
🧼 1 file clean 🧼
```

### DIFF
--- a/src/fixit/config.py
+++ b/src/fixit/config.py
@@ -231,13 +231,9 @@ def collect_rules(
                 if qualified_rule.name:
                     named_enables |= rules
                 all_rules |= rules
-            except CollectionError as e:
-                log.warning(
-                    f"Collection Error: Failed to collect enabled rules for {config.enable}: {e}"
-                )
             except Exception as e:
                 log.warning(
-                    f"Failed to collect enabled rules for {config.enable}: {e.__class__.__name__}: {e}"
+                    f"Failed to load rules '{qualified_rule.module}': {e.__class__.__name__}: {e}"
                 )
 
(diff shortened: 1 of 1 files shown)


src/fixit/config.py
```diff
@@ -226,20 +226,38 @@ def collect_rules(
...
-                named_enables |= rules
-            all_rules |= rules
+            try:
+                rules = set(find_rules(qualified_rule))
+                if qualified_rule.name:
+                    named_enables |= rules
+                all_rules |= rules
+            except CollectionError as e:
+                log.warning(
+                    f"Collection Error: Failed to collect enabled rules for {config.enable}: {e}"
```
amyreese, line 236: Log messages should only print the specific `qualified_rule` that is actively trying to be found/imported, otherwise it may be difficult to track down the specific `enable` item that triggered the warning. Same for all the warnings below.

I think a simpler message like `Failed to load rules "fixit.rules": <exception>` would be better and easier to understand. We could probably combine the two except clauses and just use the latter form for everything.

amyreese (changes requested):

amyreese (approved):

PR #480, opened 2024-08-12
Add validate command

surge119:
## Summary
Added command to Fixit to check if a config file is valid

## Test Plan
Created invalid fixit.toml files such as:
```
[tool.fixit]
enable = [
    "fixit/rules:DeprecatedABCImport",
]
disable = [
  "fixit.rules",
]

root = true
```
and ran the new command against them:
```
# Example from above
$ hatch run fixit validate-config .
Invalid config: ConfigError: invalid rule name 'fixit/rules:DeprecatedABCImport'

# Another example with invalid TOML
$ hatch run fixit validate-config .
Invalid config: TOMLDecodeError: Expected '=' after a key in a key/value pair (at line 9, column 4)
...
```

### DIFF
--- a/src/fixit/cli.py
+++ b/src/fixit/cli.py
@@ -20,8 +20,6 @@ from .rule import LintRule
 from .testing import generate_lint_rule_test_cases
 from .util import capture
 
-log = logging.getLogger(__name__)
-
 
 def splash(
     visited: Set[Path], dirty: Set[Path], autofixes: int = 0, fixed: int = 0
@@ -371,11 +369,8 @@ def validate_config(ctx: click.Context, paths: Sequence[Path]) -> None:
         from pprint import pprint  # type: ignore
 
     for path in paths:
-        path = path.resolve()
-
         try:
-            config = generate_config(path, options=options)
-            enabled = collect_rules(config)
+            collect_rules(generate_config(path.resolve(), options=options))
         except Exception as e:
             pprint(f"Invalid config: {e.__class__.__name__}: {e}")
             exit(-1)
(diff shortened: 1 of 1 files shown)


amyreese (changes requested):
I think I would prefer this to take a direct path to a config (eg, `pyproject.toml` or `fixit.toml`), read the "raw" config with `read_configs([path])`, and then manually call `parse_rule()` for each string in every enable/disable option (including in every override block). Then call `find_rules()` on each qualified rule and log errors/warnings for anything that fails to load.

That way, we're not validating the config for a single code path in the repo, but instead validating all possible rules that are referenced from the file, including in all override blocks, to make sure that rules are correctly expressed, findable, and importable.

Bonus points if the warnings/errors can reference the specific block/enable/disable section to make it clear to the user which element is causing the problem.

src/fixit/cli.py
```diff
@@ -349,3 +349,51 @@ def debug(ctx: click.Context, paths: Sequence[Path]) -> None:
...
+                try:
+                    qualified_rule = parse_rule(rule, path, configs)
+                    for _ in find_rules(qualified_rule):
+                        pass
+                except Exception as e:
+                    nonlocal exception_raised
+                    exception_raised = True
+                    pprint(
+                        f"Failed to parse rule `{rule}`: {e.__class__.__name__}: {e}"
+                    )
```
amyreese, line 382: Can we have separate try/except blocks and error messages for "parse failure" and "import failure"?

Would also be cool if `validate_rules` could take an extra context string, so that the error messages could reference what overrides block/path they were in relation to?

amyreese (commented):
Looks great overall. One last request: can we add a simple smoke test that tests this command against a bogus config file and make sure it produces some baseline expected output?  You should be able to follow examples in `tests/smoke.py` for validating expected output on stdout.

amyreese (approved):

PR #481, opened 2024-08-15
Fix validate config

surge119:
## Summary
The initial implementation would fail for overrides, since the path provided into `parse_rule` is the override path, instead of the path to the config file, which is expected.

## Test Plan
Included

### DIFF
--- a/src/fixit/config.py
+++ b/src/fixit/config.py
@@ -594,22 +594,21 @@ def generate_config(
     return config
 
 
-def validate_config(path: Path, parse_path: Optional[Path] = None) -> List[str]:
+def validate_config(path: Path) -> List[str]:
     """
     Validate the config provided. The provided path is expected to be a valid toml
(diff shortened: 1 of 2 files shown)


amyreese (commented):
I'd prefer to use the `root` terminology instead of `parse_path`.  Also, is there ever a case where `root` here wouldn't always be `path.parent`?  Can we avoid the extra argument to `validate_config` and `validate_rules`, and just set `root = path.parent` at the top of the function?

surge119:
> Also, is there ever a case where `root` here wouldn't always be `path.parent`? Can we avoid the extra argument to `validate_config` and `validate_rules`, and just set `root = path.parent` at the top of the function?

Only because of testing. Since the tests use a temporary directory, using parent for them would cause them to fail. But I suppose the tests can create new rules to reference

amyreese (approved):

PR #495, opened 2024-11-18
Replace `Optional[T]` and `Union[T, None]` with `T | None`

yangdanny97:
## Summary

The pipe syntax for unions introduced in PEP604 is more concise than `Optional`, but it only works in Python 3.10+

This PR updates the rule replacing `Union[T | None]` with `Optional[T]` to replace both with `T | None`.

## Test Plan

`make`

### DIFF
--- a/docs/guide/builtins.rst
+++ b/docs/guide/builtins.rst
@@ -728,6 +728,8 @@ Built-in Rules
     .. attribute:: AUTOFIX
         :type: Yes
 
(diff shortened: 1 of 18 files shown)


bowiechen (changes requested):
looks good! thank you for working on this. 

please fix some nits and we can merge this.

bowiechen (changes requested):

PR #496, opened 2024-11-18
Replace `typing.List` and other aliases w/ builtin equivalents after Python 3.10

yangdanny97:
## Summary

This is the inverse of UseTypesFromTyping https://github.com/Instagram/Fixit/blob/main/src/fixit/rules/use_types_from_typing.py

In Python versions after 3.10, we can replace `typing.List` and other aliases with the builtin equivalents.

## Test Plan

`make`

closes https://github.com/Instagram/Fixit/pull/482


yangdanny97:
Oh, I just realized this is almost a duplicate of https://github.com/Instagram/Fixit/pull/482

This one adds `typing.Type` support which the other one doesn't have, but I'll let the maintainers decide which to move forward with.

PR #502, opened 2025-01-09
Drop testing on 3.8, test on 3.13

amyreese:


zsol (approved):
There are some fun failures in 3.13, but otherwise LGTM

PR #503, opened 2025-01-20
Allow dash in rule names

zsol:
Since rules are dynamically imported, they don't actually have to adhere to identifier naming.

This allows putting rules in directories with `-` in them.


amyreese:
gross but ok 😅

PR #506, opened 2025-03-26
Pass sitepackages and python version as a flag

MaggieMoss:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,3 +44,27 @@ jobs:
           python -m hatch run docs:build
       - name: Generated changes
         run: git diff --exit-code
+  pyrefly:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.12"]
+        os: [macOS-latest, ubuntu-latest, windows-latest]
+    steps:
+      - name: Checkout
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -44,3 +44,27 @@ jobs:
...
+      - name: Install
+        run: |
+          python -m pip install -U pip hatch
+          python -m hatch env create
+          python -m hatch env create lint
+      - name: Set SITE_PACKAGE_PATH
+        run: echo "SITE_PACKAGE_PATH=$(python -m site --user-site)" >> $GITHUB_ENV
+      - name: Typecheck
+        run: |
+          python -m hatch run pyrefly_check --python_version=${{matrix.python-version}} --site-package-path=${{env.SITE_PACKAGE_PATH}}
```
zsol, line 70: Two concerns with this:
1. I'm surprised it works on Windows `$(...)` is a bash-ism
2. This is getting a bit complex, so maybe worth just wrapping the pyrefly invocation into a trivial python script that already knows its own version and site-package path?

MaggieMoss: cc: @connernilsen

MaggieMoss: Conner is working on automatically detecting this, I'll close this PR until that is complete and released

PR #552, opened 2026-02-23
Exit with non-zero when autofix not applied

### LINKED ISSUE #258: Fixit cli doesn't return non-zero exit code upon exceptions - add --exit-non-zero / --exit-zero options
Hello,

Commands like

```shell
python -m fixit.cli.apply_fix
```

can fail for some files in the CI silently and there is no way to detect it via the exit code - parsing also cannot be done reliably:

```
Encountered the following error while parsing source code in file /home/runner/work/xxx.py:
Syntax Error @ 59:19.
Incomplete input. Encountered 'operation', but expected ';', or 'NEWLINE'.
```

Linters like `flake8` exit with non-zero code per default and they have an option like `--exit-zero` to ignore errors. Formatter `black` has option `--check`.

I would suggest to make `fixit` fail is some exception happens per default and add `--exit-zero` to opt out. Otherwise one can make this an opt-in functionality like `--exit-non-zero`.

I think this is quite important, especially after LibCST parser update, because problems can easily go unnoticed, if the is no way to make `fixit` return a non-zero code upon errors.

Thanks!

### LINKED ISSUE #409: `fixit fix` exits with  `0` even if there were errors
`fixit fix` will exit with an exit code of `0` if it finds errors but does not apply fixes. This makes using it in CI (or pre-commit hooks) awkward because it hides errors that can't be auto-fixed. The solution is to run the tool *twice: once to fix fixable errors, then once again to re-report them in a way that can cause a Ci job to fail. This is a frustrating duplication of work.

Given the outcome of #258, I think this is a bug rather than intentional design, but I'm happy to be corrected if there's another change I've missed 🙂 

There are two cases where I would expect a non-zero exit code from `fixit fix`.

## failure to run

When it fails to evaluate the rules at all because of an error:

```console
$ cat t.py
this is not valid python

$ fixit fix -a t.py
t.py: EXCEPTION: Syntax Error @ 1:1.
parser error: error at 1:24: expected one of !=, %, &, (, *, **, +, ,, -, ., /, //, ;, <, <<, <=, ==, >, >=, >>, @, NEWLINE, [, ^, and, if, in, is, not, or, |

this is not valid python
^
Traceback (most recent call last):
  File "/private/tmp/tmpvenv-c329d/venv/lib/python3.11/site-packages/fixit/api.py", line 100, in fixit_bytes
    runner = LintRunner(path, content)
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/private/tmp/tmpvenv-c329d/venv/lib/python3.11/site-packages/fixit/engine.py", line 55, in __init__
    self.module: Module = parse_module(source)
                          ^^^^^^^^^^^^^^^^^^^^
  File "/private/tmp/tmpvenv-c329d/venv/lib/python3.11/site-packages/libcst/_parser/entrypoints.py", line 109, in parse_module
    result = _parse(
             ^^^^^^^
  File "/private/tmp/tmpvenv-c329d/venv/lib/python3.11/site-packages/libcst/_parser/entrypoints.py", line 55, in _parse
    return parse(source_str)
           ^^^^^^^^^^^^^^^^^
libcst._exceptions.ParserSyntaxError: Syntax Error @ 1:1.
parser error: error at 1:24: expected one of !=, %, &, (, *, **, +, ,, -, ., /, //, ;, <, <<, <=, ==, >, >=, >>, @, NEWLINE, [, ^, and, if, in, is, not, or, |

this is not valid python
^
🛠️  1 file checked, 1 file with errors 🛠️

$ echo $?
0
```

I would expect `fixit` to exit with a non-zero code in the above case.

## errors found but not fixed

`fixit fix` also exits `0` if errors were found but no fixes were applied:

```console
$ cat t.py
try:
    ...
except ValueError or KeyError:
    ...

$ fixit fix -a t.py
t.py@1:0 AvoidOrInExcept: Avoid using 'or' in an except block. For example:'except ValueError or TypeError' only catches 'ValueError'. Instead, use parentheses, 'except (ValueError, TypeError)'
🛠️  1 file checked, 1 file with errors 🛠️

$ echo $?
0
```

I would also expect it to exit non-zero in this case, although the suggestion in #258 for that to be opt-in seems reasonable if backwards-compatibility is a concern.

---

Fixit version 2.1.0
Python 3.11.5 (main, Aug 24 2023, 15:09:45) [Clang 14.0.3 (clang-1403.0.22.14.1)]

soraxas:
## Summary

Closes #409 

This PR improves the exit code handling in the `fix` command when multiple issues are encountered in a single run.

* Refactored the `fix` command in `src/fixit/cli.py` to set exit codes more accurately: now, exit code `2` is set for syntax errors, exit code `1` for unfixable violations, and both are combined to 3 if both types of issues are present. This ensures that the CLI communicates the exact result of the operation.
* This is in accordance to https://github.com/Instagram/Fixit/commit/5e49b17a135b807ac03ab130cb4c782f9609e13d
    > Return non-zero exit codes for violations/errors
	> 
	> This tracks exit codes based on the existence of violations and/or
	> errors when linting files.
	> 
	> - 0: everything clean
	> - 1: violations only
	> - 2: errors only
	> - 3: both violations and errors
	> 
	> Fixes https://github.com/Instagram/Fixit/issues/258
	> 
	> ghstack-source-id: https://github.com/Instagram/Fixit/commit/2c18d7a06f67313d29863e55a6cb45a97c655146
	> Pull Request resolved: https://github.com/Instagram/Fixit/pull/283

## Test Plan

* Added a new test method `test_fix_exit_codes` to verify the exit codes for unfixable violations, syntax errors, and scenarios where both occur together.

### DIFF
--- a/src/fixit/cli.py
+++ b/src/fixit/cli.py
@@ -218,27 +218,44 @@ def fix(
         visited.add(result.path)
         # for STDIN, we need STDOUT to equal the fixed content, so
         # move everything else to STDERR
-        if print_result(
+        is_dirty = print_result(
             result,
             show_diff=interactive or diff,
             stderr=is_stdin,
             output_format=config.output_format,
             output_template=config.output_template,
-        ):
+        )
+        if is_dirty:
             dirty.add(result.path)
-            if autofix and result.violation and result.violation.autofixable:
-                autofixes += 1
-                fixed += 1
-        if interactive and result.violation and result.violation.autofixable:
+
+        if result.error:
+            exit_code |= 2
+            continue
+
+        violation = result.violation
+        if not violation:
+            continue
+
+        violation_fixed = False
+        if violation.autofixable:
             autofixes += 1
-            answer = click.prompt(
-                "Apply autofix?", default="y", type=click.Choice("ynq", False)
-            )
-            if answer == "y":
-                generator.respond(True)  # noqa: B038
+            if autofix:
                 fixed += 1
-            elif answer == "q":
-                break
+                violation_fixed = True
+            elif interactive:
+                answer = click.prompt(
+                    "Apply autofix?", default="y", type=click.Choice("ynq", False)
+                )
+                if answer == "y":
+                    generator.respond(True)  # noqa: B038
+                    fixed += 1
+                    violation_fixed = True
+                elif answer == "q":
+                    exit_code |= 1
+                    break
+
+        if not violation_fixed:
+            exit_code |= 1
 
     splash(visited, dirty, autofixes, fixed)
     ctx.exit(exit_code)
--- a/src/fixit/tests/smoke.py
+++ b/src/fixit/tests/smoke.py
@@ -209,6 +209,53 @@ class SmokeTest(TestCase):
             self.assertIn("broken.py: EXCEPTION: Syntax Error @ 1:", result.output)
             self.assertEqual(result.exit_code, 3)
 
+    def test_fix_exit_codes(self) -> None:
+        with self.subTest("unfixable violation"):
+            with TemporaryDirectory() as td:
+                tdp = Path(td).resolve()
+                dirty = tdp / "dirty.py"
+                dirty.write_text(
+                    "try:\n    pass\nexcept ValueError or KeyError:\n    pass\n"
+                )
+
+                result = self.runner.invoke(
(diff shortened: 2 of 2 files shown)


soraxas:
Hi @amyreese, any update?

PR #566, opened 2026-06-02
Remove mypy

stroxler:
I don't believe mypy is actually used - at this point we type check with Pyrefly

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,6 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
-          - "3.9"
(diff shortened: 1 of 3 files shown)


stroxler:
Looks like mypy is still wired into the setup via tests - I had noted that it isn't a separate CI step, which is true, but there's some unusual indirection going on.

I'll think on this a bit, my instinct is that it's not worth having two type checkers, but I don't have strong opinions

amyreese:
You just need to remove the call to mypy from the makefile here: https://github.com/Instagram/Fixit/blob/main/makefile#L28

PR #571, opened 2026-08-25
Batch Fixit traversal - 11% improvement

tpolasek:
Resolve metadata once for the active rule batch and pre-index visitor dispatch. Collect unused suppressions during the same CST walk and finalize them after report filtering.

## Test Plan
Add coverage for inherited visitor discovery and used and unused suppressions.


stroxler (changes requested):
I think this one needs a rebase?

